PR #3, opened 2021-09-06
Fix homepage reuses

maudetes:
Fix etalab/data.gouv.fr#337.

See previous discussions at https://github.com/etalab/udata-gouvfr/pull/628.


abulte (approved):
Back to the future!

PR #5, opened 2021-09-09
Pin setuptools==53.0.0

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/508.

Sadly, the [patch](https://github.com/pypa/setuptools/issues/2775) didn't fix our issue, because we do use a lib that uses `use_2to3=True`.

We'll need to update our dependencies that rely on `use_2to3=True` in the future (https://github.com/etalab/data.gouv.fr/issues/512).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,8 +22,10 @@ jobs:
       - run:
           name: Create venv and install pip-tools
           command: |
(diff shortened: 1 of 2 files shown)


maudetes:
Not needed anymore due to #6.

PR #10, opened 2021-09-16
Add read only banner

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/468.

Based on https://github.com/etalab/udata-gouvfr/pull/509 and https://github.com/opendatateam/udata/pull/2565/files.

Some modifications had been lost during *refonte*.

### DIFF
--- a/theme/js/config.js
+++ b/theme/js/config.js
@@ -159,6 +159,11 @@ export const dataset_max_resources_uncollapsed = _jsonMeta(
  */
 export const markdown = _jsonMeta("markdown-config");
 
+/**
+ * Whether the 'read only mode' feature is enabled or not.
+ */
+ export const read_only_enabled = _jsonMeta('read-only-enabled');
+
 // New generic `js-config-*` variables : simply add them to `metadata.html` with a meta name="js-config-something", content="yourValue" and import this file.
 // Everything is exported in the `values` variable as key: value pairs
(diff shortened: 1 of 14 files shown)


theme/js/config.js
```diff
@@ -200,4 +205,5 @@ export default {
   values,
   dataset_max_resources_uncollapsed,
   markdown,
+  read_only_enabled
```
abulte, line 208:
```suggestion
  read_only_enabled,
```

udata_front/theme/gouvfr/templates/header.html
```diff
@@ -1,5 +1,13 @@
 {% set banner_activated = config['BANNER_ACTIVATED'] %}
+{% set read_only_mode = config['READ_ONLY_MODE'] %}
 {% set next_url = url_for(request.endpoint, **request.view_args) if not request.routing_exception else url_for('site.home') %}
+
+{% if read_only_mode %}
+{% block read_only_banner %}
```
abulte, line 6: Why do we need a block here? 🤔 An include should be enough.

abulte (approved):

udata_front/theme/gouvfr/templates/header.html
maudetes: Cause I'm a copycat :smile_cat: !

PR #11, opened 2021-09-21
Unset vue delimiters

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/495.

We actually don't need to modify existing `{{` delimiters in `.vue` components as they are not impacted by this.

I changed for `v-text` and then rolled back the changes, but if cleaner we can get back to using `v-text` (what do you think @gmougeolle?).

I currently use a deprecated delimiter config (in 3.1.0: https://v3.vuejs.org/api/options-misc.html#delimiters), but couldn't make `compilerOptions.delimiters` work yet.


gmougeolle:
We might have to check that it still works that way with Vue 3.2, but right now it looks perfect for us : no more risks on the HTML templates, and cleaner `.vue`files !

maudetes:
Yes, it seems to work as well in vue 3.2 :)

image

PR #12, opened 2021-09-21
Add a Request Membership action on organization page

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/438.

Used a simple prompt. See the following:

image
image
image
image

For now, the pending status is not updated dynamically.

### DIFF
--- a/theme/js/components/organization/request-membership.vue
+++ b/theme/js/components/organization/request-membership.vue
@@ -37,13 +37,11 @@ data() {
       this.$api.post("organizations/" + this.orga + "/membership/", {comment: this.comment})
           .then(data => {
             alert(this.$t('A request has been sent to the administrators'));
+            window.location.reload();
         })
         .catch(error => {
             alert(this.$t('Error while requesting membership'));
             log.error(error);
-        })
-        .finally(()=> {
-            window.location.reload();
         });
       }
   }
--- a/theme/js/components/organization/request-membership.vue
+++ b/theme/js/components/organization/request-membership.vue
@@ -32,7 +32,7 @@ data() {
 
       this.$auth(this.$t("You must be connected to join an organization."));
 
-      this.comment = prompt(this.$t('You can add some details here for your membership request.'))
+      this.comment = prompt(this.$t('You can add some details here for your membership request.'));
 
       this.$api.post("organizations/" + this.orga + "/membership/", {comment: this.comment})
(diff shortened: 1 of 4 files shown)


theme/js/components/organization/request-membership.vue
```diff
@@ -0,0 +1,48 @@
...
+
+# Request Membership
+
+A simple request membership prompt.
+-->
+
+<template>
+    <li class="my-md">
+        <a @click.prevent="JoinOrga"
+      class="nav-link">{{ $t('Ask to join the organization as a producer') }}</a>
```
abulte, line 15: Was that the previous wording?

abulte (commented):
Since we're going full vintage ™️ maybe reload the page after submit to update the request status? Might be violent though :godmode:

theme/js/components/organization/request-membership.vue
maudetes: Nope, it used to be "join/rejoindre" only in legacy. I used the suggestion in https://github.com/etalab/data.gouv.fr/issues/438#issuecomment-915266108.

theme/js/components/organization/request-membership.vue
```diff
@@ -36,12 +36,15 @@ data() {
...
-              alert(this.$t('Error while requesting membership'));
-              log.error(error);
-          });
+            alert(this.$t('A request has been sent to the administrators'));
+        })
+        .catch(error => {
+            alert(this.$t('Error while requesting membership'));
+            log.error(error);
+        })
+        .finally(()=> {
```
abulte, line 45: do we want to reload when there's an error?

maudetes: Nope, indeed!

theme/js/components/organization/request-membership.vue
```diff
@@ -0,0 +1,49 @@
...
+    return {
+        comment: '',
+    };
+},
+  methods: {
+    JoinOrga: function () {
+
+      this.$auth(this.$t("You must be connected to join an organization."));
+
+      this.comment = prompt(this.$t('You can add some details here for your membership request.'))
```
abulte, line 35:
```suggestion
      this.comment = prompt(this.$t('You can add some details here for your membership request.'));
```

abulte (approved):

PR #19, opened 2021-10-05
fix: Modals now properly trigger

gmougeolle:


maudetes (approved):
Thank you !
Please modify the existing entry with this PR (https://github.com/etalab/udata-front/blob/bc30a9123134033067adfe3557c70ee2e45180cd/CHANGELOG.md#L11) in the changelog :)

PR #20, opened 2021-10-05
Remove PDF link to LOv2 and link to GitHub markdown

abulte:
Fix https://github.com/etalab/data.gouv.fr/issues/545

Not sure if we need a CHANGELOG entry?


udata_front/theme/gouvfr/templates/footer.html
```diff
@@ -57,9 +57,9 @@ <h4>Réseaux sociaux</h4>
                         </li>
                     </ul>
                     <p class="text-black fs-sm">
-                        Sauf mention contraire, tous les contenus de ce site sont sous licence
-                        <a href="https://www.etalab.gouv.fr/wp-content/uploads/2017/04/ETALAB-Licence-Ouverte-v2.0.pdf"
-                            title="">etalab-2.0</a>
+                        Sauf mention contraire, tous les contenus de ce site sont sous
```
maudetes, line 60: And now I see that the text is in French and we don't have any translation! Could you add it please :pray: ?

maudetes (commented):
Ok, let's not add a changelog entry

PR #21, opened 2021-10-06
feat: add roles to block of content

nicolaskempf57:
Add WAI-ARIA roles to header, navigations, search and main content.
Add `<main>` tag around base template main content.

Close https://github.com/etalab/data.gouv.fr/issues/544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Unset vue delimiters used in html templates to prevent injections [#11](https://github.com/etalab/udata-front/pull/11)
 - Fix temporal coverage order in search results metadata [#14](https://github.com/etalab/udata-front/pull/14)
 - VueJS multiple mount points with a global event bus [#15](https://github.com/etalab/udata-front/pull/15) [#19](https://github.com/etalab/udata-front/pull/19)
-- Fix RGAA criterion 12.6 Block of contents from multiple pages can be reched or skipped [#21](https://github.com/etalab/udata-front/pull/21)
+- Fix RGAA criterion 12.6 Block of contents from multiple pages can be reached or skipped [#21](https://github.com/etalab/udata-front/pull/21)
 
 ## 1.0.0 (2021-09-16)
 


udata_front/theme/gouvfr/templates/header.html
```diff
@@ -29,7 +29,10 @@
                             {% include theme('svg/marianne.svg') %}
                         </a>
                     </div>
-                    <nav class="col hidden visible-xl mobile-menu row-inline justify-end self-start vuejs">
+                    <nav
+                            class="col hidden visible-xl mobile-menu row-inline justify-end self-start vuejs"
+                            role="navigation"
+                    >
                         <a href="/datasets" title="{{ _('Search for data') }}" class="mr-md"
```
maudetes, line 36: Shouldn't there be a `role="search"` here? Shouldn't the `role="navigation"` be on the menu part only?

udata_front/theme/gouvfr/templates/base.html
```diff
@@ -14,7 +14,9 @@
     {% endif %}
 </div>
 {% endwith %}
-{% block content %}{% endblock %}
+<main>
```
maudetes, line 17: We don't need this tag for accessibility right?

udata_front/theme/gouvfr/templates/header.html
```diff
@@ -78,15 +81,15 @@
         <nav class="header-nav">
             <div class="hagrid">
                 <div class="row justify-between align-items-center">
-                    <nav class="nav-tabs col">
+                    <nav class="nav-tabs col" role="navigation">
```
maudetes, line 84: Not sure if this is an issue, but on mobile, the navigation role appears twice.

udata_front/theme/gouvfr/templates/base.html
nicolaskempf57: `<main>` or a tag with `role="main"` is required on each page

udata_front/theme/gouvfr/templates/header.html
nicolaskempf57: We can have multiple navigations per page but we may want to add an aria-label to them (as stated here https://disic.github.io/guide-integrateur/1-gabarit-general.html#html5aria)

maudetes: ah yes indeed, nice ! :+1:

nicolaskempf57:
I added `role = navigation` back to pagination because it's a navigation based on https://disic.github.io/guide-integrateur/2-navigation.html#collectionpages

maudetes:
> I added `role = navigation` back to pagination because it's a navigation based on https://disic.github.io/guide-integrateur/2-navigation.html#collectionpages

ah yes indeed, thank you!

udata_front/theme/gouvfr/templates/header.html
maudetes: After discussion, it seems that since hidden elements are removed from the accessibility tree, we can have the navigation and search roles twice for desktop and responsive.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Unset vue delimiters used in html templates to prevent injections [#11](https://github.com/etalab/udata-front/pull/11)
 - Fix temporal coverage order in search results metadata [#14](https://github.com/etalab/udata-front/pull/14)
 - VueJS multiple mount points with a global event bus [#15](https://github.com/etalab/udata-front/pull/15) [#19](https://github.com/etalab/udata-front/pull/19)
+- Fix RGAA criterion 12.6 Block of contents from multiple pages can be reched or skipped [#21](https://github.com/etalab/udata-front/pull/21)
```
maudetes, line 12:
```suggestion
- Fix RGAA criterion 12.6 Block of contents from multiple pages can be reached or skipped [#21](https://github.com/etalab/udata-front/pull/21)
```

maudetes (approved):
I think we're good :)

PR #26, opened 2021-10-21
Fix feedback on dataset quickfixes

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/582

### DIFF
--- a/udata_front/theme/gouvfr/translations/gouvfr.pot
+++ b/udata_front/theme/gouvfr/translations/gouvfr.pot
@@ -8,8 +8,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: udata-front 1.1.1.dev0\n"
 "Report-Msgid-Bugs-To: data.gouv@data.gouv.fr\n"
-"POT-Creation-Date: 2021-10-15 09:41+0200\n"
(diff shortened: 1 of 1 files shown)


udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -179,6 +183,10 @@ <h2 class="subtitle">{{ _('Informations') }}</h2>
                                  <dd>
                                      {{dataset.id}}
                                  </dd>
+                                {% if external_source %}
+                                <dt class="f-bold text-grey-400">{{ _('Remote source') }}</dt>
```
maudetes, line 187: Why this styling? I think we can keep the `dt` style only?

maudetes (commented):
Thank you for these quick fixes (they deserve their name better than the dataset navigation quickfixes) ! It looks (almost) good for me :pray:

Additionally, did you try to take a look at the embed font size?

udata_front/theme/gouvfr/templates/dataset/display.html
nicolaskempf57: My bad, it came from the old sources

PR #28, opened 2021-10-21
Override oauth route to serve udata-front template

maudetes:
Related to https://github.com/etalab/data.gouv.fr/issues/466.

This is a proposal to override the `/authorize` route in `udata-front` to render the udata-front template if the plugin is loaded.

It didn't appear to be necessary in our case, but we could look to make sure to unregister the udata `/authorize` route ([example from stackoverflow](https://stackoverflow.com/a/49830875)).


maudetes:
Used hooks instead in https://github.com/etalab/udata-front/pull/29.

PR #33, opened 2021-10-28
Old UI kit review

gmougeolle:
I did a bit of fixing for our Stylemark build : it now correctly includes JS/Vue files and correctly identifies examples' originating file.

I also did a review of all existing styles in order for us to decide what to do with some of them : 

- Primary button
Capture d’écran 2021-10-28 à 16 08 35

This thing is used in many places but not any important ones : preview modal, schema modal, oauth form, schema modal button, top of the page error section (`base.html`), register / login form and "sysadmin" button.
**It exists in 18+ shades but is only used in its default color** 😅 
I think we can easily delete the colored variants and even start replacing it in older templates with the DSFR button if we want to but the style is quite different.

👉  https://gouvfr.atlassian.net/wiki/spaces/DB/pages/217284660/Boutons+-+Buttons

- Secondary button

Capture d’écran 2021-10-28 à 16 08 40

This button is actually used way more than the primary button, hence why we can say it was a good naming strategy we went after. This would be way harder to replace in its current state, but some colors can probably be pruned : used one seem to only include `green-300`, `blue-400`, `grey-(200|400|500)`, `orange-100` and `white`.

- Action button

Capture d’écran 2021-10-28 à 16 08 50

This is an oldie : this kind of button slowly got replaced to the point where it's only used in the ressource card, reuse and dataset pages for the "add a reuse / dataset / discussion" links. This wouldn't be too much work to replace, but is still being used in important places so it probably needs some design thinking beforehand.

- Dropdown

Capture d’écran 2021-10-28 à 16 09 11

Only used in two places : to sort threads and in the footer to select your language. Very hard to use bit of UI and very rarely used, we can probably replace that with a DSFR version easily.

👉  https://gouvfr.atlassian.net/wiki/spaces/DB/pages/223019306/Liste+d+roulante+-+Select

- Icon size

This thing doesn't have a screenshot because it doesn't really do anything more than set a width and height to svg classes. Weird thing is it only exists in 30x30px (`icon-lg`) and is only used on the "certified" icon in the organization display page. DSFR has 3 sizes available, `md` (16x16), `lg` (24x24) and `xl`(32x32) that could maybe replace those utilities but should be tested beforehand as we're using SVG icons.

👉  https://gouvfr.atlassian.net/wiki/spaces/DB/pages/222331396/Ic+nes+-+Icons

- Tab-panel

Capture d’écran 2021-10-28 à 16 09 50

This is only used in the reuses pages now and can safely be marked as "depreciated" until the reuse page gets redone.

- Tags

Capture d’écran 2021-10-28 à 16 10 02

Those are used in two or three places (dataset description, reuse description, some error/state messages) in two different sizes (normal and big). DSFR has a much more complete system but the style is also a bit different : 

👉  https://gouvfr.atlassian.net/wiki/spaces/DB/pages/310706305/Tag

- Callouts

Capture d’écran 2021-10-28 à 16 10 08

This is not used anymore and can safely be deleted 🤠 

- Aileron font

This is now Marianne but the doc is outdated, I'm going to replace it

Capture d’écran 2021-10-28 à 16 10 25

- Colors

Pretty colors. WIP : knowing which is used, which isn't. What to do with those knowing that DSFR comes with others ?

Capture d’écran 2021-10-28 à 16 36 03

**Text colors**

Unused text colors :

- text-blue-100
- text-blue-300
- text-yellow-100
- text-grey-50
- text-grey-100
- text-grey-200
- text-grey-380
- text-green-500

Unused bg colors : 

- bg-blue-200
- bg-blue-500
- bg-yellow-100
- bg-grey-100
- bg-grey-200
- bg-grey-300
- bg-green-400
- bg-green-500

Unused variables : 

- @black
- @yellow-100
- @green-500

Deleting unused text or bg color classes by themselves is not very interesting, however deleting whole colors can be useful. It looks like `yellow-100` and `green-500` can safely be deleted.

- Interactions

Those are JS docs that just showed up in Stylemark.
Tabs are used on the home page (and probably somewhere else) and are pretty simple and can probably be kept as-is.
Clipboard is very simple as well and can be kept too. Accordions too.
Modals probably need some refactoring as it was an afterthought and they may not correspond to what we need.


maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/484.

maudetes (approved):

maudetes:
Ah could you add an entry in the changelog? :)

PR #34, opened 2021-11-04
Update oauth page style

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/466
Based on https://www.figma.com/file/kfpGHDywdXEr0phdHfjTMc/Live%2FFixes?node-id=437%3A11198

Capture d’écran 2021-11-04 à 15 32 53

Because of oauth client and organization image ratio, we might want to change the header.
See below for some examples.

Capture d’écran 2021-11-04 à 15 22 15
Capture d’écran 2021-11-04 à 15 22 37

### DIFF
--- a/udata_front/views/gouvfr.py
+++ b/udata_front/views/gouvfr.py
@@ -132,10 +132,16 @@ def dataset_apis(ctx):
     return theme.render('dataset-apis.html', apis=dataset.extras.get(APIGOUVFR_EXTRAS_KEY))
 
 
-@template_hook('oauth_theme_content')
+@template_hook('oauth_authorize_theme_content')
 def oauth_theme_content(ctx):
     grant = ctx['grant']
     return theme.render('api/oauth_authorize.html', grant=grant)
+
+
+@template_hook('oauth_error_theme_content')
+def oauth_theme_content(ctx):
+    request = ctx['request']
+    return theme.render('api/oauth_error.html', error=request.args.get('error'))
 
 
 # TODO : better this, redirect is not the best. How to serve it instead ?!
--- a/udata_front/views/gouvfr.py
+++ b/udata_front/views/gouvfr.py
@@ -133,13 +133,13 @@ def dataset_apis(ctx):
 
 
 @template_hook('oauth_authorize_theme_content')
-def oauth_theme_content(ctx):
+def oauth_authorize_theme_content(ctx):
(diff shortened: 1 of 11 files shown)


maudetes (commented):
Nice!

Indeed, the orga and service image ratio is confusing right now. Should we ask for a mock-up update?

Let's not forget to extract additional translations before merging.

nicolaskempf57:
> Indeed, the orga and service image ratio is confusing right now. Should we ask for a mock-up update?

We can limit the height in the meantime but the logos will be difficult to read. We can still merge this PR before the redesign because right now the page is really confusing and logo are a broader problem and look off (even on an [organization page](https://www.data.gouv.fr/fr/organizations/prefecture-des-landes/)).

maudetes:
> We can limit the height in the meantime but the logos will be difficult to read. We can still merge this PR before the redesign because right now the page is really confusing and logo are a broader problem and look off (even on an [organization page](https://www.data.gouv.fr/fr/organizations/prefecture-des-landes/)).

Yes agreed, let's add a height limit for now and investigate on logo later.

nicolaskempf57:
There is now a max-height ! I also fix a margin issue when there was no image at all (no organization + no client image).

Capture d’écran 2021-11-09 à 13 52 26
Capture d’écran 2021-11-09 à 13 54 38

maudetes (approved):
:tada:

udata_front/views/gouvfr.py
```diff
@@ -132,12 +132,18 @@ def dataset_apis(ctx):
...
+@template_hook('oauth_authorize_theme_content')
 def oauth_theme_content(ctx):
     grant = ctx['grant']
     return theme.render('api/oauth_authorize.html', grant=grant)
 
 
+@template_hook('oauth_error_theme_content')
+def oauth_theme_content(ctx):
+    request = ctx['request']
+    return theme.render('api/oauth_error.html', error=request.args.get('error'))
```
maudetes, line 144: Could you test with an actual error?

nicolaskempf57: I test it by manually creating a URL with an error like `/en/oauth/error?error=Incorrect%20field` but I don't know how to make an actual oauth error going on this page. When I made an error with authorize parameters, the error is displayed in an empty page, not this template.

PR #41, opened 2021-11-19
Standardize article discussions

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/578
Close https://github.com/etalab/data.gouv.fr/issues/649
Close https://github.com/etalab/data.gouv.fr/issues/650

This PR is the last one regarding discussions. It fixes some inconsistencies on article page.
Also, it adds not-so-quick fixes to discussions as seen on the parent issue : https://github.com/etalab/data.gouv.fr/issues/578#issuecomment-964230179

There are some inconsistencies between the design and the actual implementation :
- permalinks were missing on Figma
- reply were shown on each reply  on Figma but you can only reply to a thread
- reply form wasn't designed. I use a raw DSFR textarea. We can remove the bottom border to match the new select too.
Capture d’écran 2021-11-19 à 15 50 10

- single thread view wasn't design so the "green well" is still here.
Capture d’écran 2021-11-19 à 15 49 52

- permalinks were only an icon, it's a bad practice so they are replace with text + icon

The new select is a DSFR select without bottom border, so we're not really using the DSFR as it should be.

The previous design used a `.nav-link`(an underlined text with a `>` on the left). I keep it as is so it had the same problem as our buttons where icons are underlined. We can reuse this fix for other cases :
-  remove the button underline
- add our content in a span
- add our underline rule to the span (on hover for buttons)

If we want to change the link style,  we can use a simple DSFR link with our new blue when we implement it.


udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -402,16 +402,13 @@ <h2 id="community-reuses" class="fr-h2">{{ _('Reuses') }} <sup>{{ reuses|length
...
-        </div>
         <div class="vuejs">
-            <discussion-threads ref="discussions" subject-id="{{ dataset.id }}"
-                subject-class="{{ dataset.__class__.__name__ }}">
+            <discussion-threads
+                ref="discussions"
+                subject-id="{{ dataset.id }}"
+                subject-class="{{ dataset.__class__.__name__ }}"
+            >
+                <p>{{ _('Discussion between the organization and the community about this dataset.') }}</p>
```
maudetes, line 411: Maybe change wording to `producer` or make it conditional to include user as well?

maudetes: Could you also update the wording accordingly here (and change `dataset` by `reuse`)? https://github.com/etalab/udata-front/blob/2054d776d3bd6f3901497e354cc7d75233733cc0/udata_front/theme/gouvfr/templates/reuse/display.html#L241 :pray:

theme/less/components/discussions.less
```diff
@@ -217,125 +217,18 @@ You can use a discussion creation button too.
```
maudetes, line 217: I think you need to update the discussions stylemark examples.

maudetes (commented):
Thank you for this PR and the clear explanations!

Some global comments:
* The issue probably appeared when paginating resources, but when opening a permalink of a particular discussion (ex: [this discussion](http://dev.local:7000/en/datasets/open-damir-base-complete-sur-les-depenses-dassurance-maladie-inter-regimes/#discussion-6151c7a7953d864261ce2e9b)), the scroll is not on the discussion but above?
* Some contrast issues on closed threads, ex: http://dev.local:7000/en/datasets/jours-feries-en-france/#community-discussions.
* There could be a padding between discussions and pagination
image
* This bug existed before this PR, but could we add a required attribute to prevent the error message if text/title empty?

theme/js/components/discussions/threads-create.vue
```diff
@@ -1,50 +1,53 @@
...
+      v-if="showCreateButton && !showForm"
+      :onClick="displayForm"
     >
-      <span v-html="AddIcon"></span>
-      <span>{{ $t("Start a new discussion") }}</span>
-    </a>
-    <div v-if="showForm" class="thread-wrapper">
-      <div class="thread-header">
-        <div class="thread-title">{{ $t("New discussion") }}</div>
+    </ThreadsCreateButton>
```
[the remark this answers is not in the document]
nicolaskempf57: This button is here to allow the same use case as before : a button becoming the form on click. This is not currently used so we can remove it if we want to.

theme/js/components/discussions/thread.vue
```diff
@@ -1,72 +1,66 @@
...
-          <div class="thread-box">
-            <p class="m-0">{{ comment.content }}</p>
           </div>
         </article>
       </transition-group>
-      <article
-        class="thread-collapse"
-        v-if="_collapsed"
+      <div class="fr-px-3w" v-if="_collapsed">
+        <button
```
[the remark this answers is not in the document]
nicolaskempf57: I'll try to keep the same clickable area. I made some modifications here to make it accessible.

theme/js/components/discussions/threads.vue
```diff
@@ -222,19 +236,21 @@ export default {
     },
     // Can be called from outside the component to start a new thread programmatically
```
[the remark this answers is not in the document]
nicolaskempf57: I kept the same because both have the same use case but we can move these to a stylemark documentation and update them when https://github.com/etalab/udata-front/pull/33 is merged.

theme/js/components/discussions/thread.vue
```diff
@@ -80,62 +74,59 @@
...
     },
   },
-  components: {
-    "thread-reply": ThreadReply,
-    Avatar,
-    Author,
-  },
   methods: {
-    discussionUrl: (id, link = false) => (link ? "#" : "") + "discussion-" + id, //Permalink helpers
-    commentUrl: function (id, index, link = false) {
```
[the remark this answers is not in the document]
nicolaskempf57: When talking about the missing elements from Figma, we decided to remove it because there is no visual difference between a discussion permalink and a comment permalink :
- the message is the same
- the page scroll at the top of the discussion
- there is no visual information about the comment linked

But we can kept them as is or update them.

nicolaskempf57:
> * [ ]  The issue probably appeared when paginating resources, but when opening a permalink of a particular discussion (ex: [this discussion](http://dev.local:7000/en/datasets/open-damir-base-complete-sur-les-depenses-dassurance-maladie-inter-regimes/#discussion-6151c7a7953d864261ce2e9b)), the scroll is not on the discussion but above?

We can update the timeout before scrolling to a particular discussion but it's only a workaround and that doesn't fix the causes. A more stable solution would be to use our new event bus to dispatch events when our resources are displayed and wait for this to do the scroll.

> * [ ]  Some contrast issues on closed threads, ex: http://dev.local:7000/en/datasets/jours-feries-en-france/#community-discussions.

I update the `text-blue-200` to a `<strong>` tag. We can use another color instead but this is not clickable.

theme/js/components/discussions/threads.vue
```diff
@@ -1,66 +1,71 @@
...
+      <template v-if="loading" key="loader">
         <Loader class="mt-md" />
-      </div>
-      <div v-else>
+      </template>
+      <template v-else>
         <div v-if="threadFromURL">
-          <div class="well well-secondary-green-300">
-            <div class="row-inline justify-between">
+          <div class="fr-px-3w well well-secondary-green-300">
```
nicolaskempf57, line 35: Do we want to try a [DSFR callout](https://gouvfr.atlassian.net/wiki/spaces/DB/pages/222331196/Mise+en+avant+-+Call+out) here ? The main difference is the button which is only a "close" for us (and only a `x` before) but is a real button on the DSFR callout.

theme/less/components/discussions.less
nicolaskempf57: True ! We can also move it back to vue files when https://github.com/etalab/udata-front/pull/33 is merged

theme/js/components/discussions/threads.vue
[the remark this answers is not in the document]
maudetes: Maybe we could update the line to add the scroll into view idea:
```
Can be called from outside the component to start a new thread programmatically and scroll into view
```

theme/js/components/discussions/threads.vue
maudetes: Yes please, I'm curious! :pray:

maudetes:
> We can update the timeout before scrolling to a particular discussion but it's only a workaround and that doesn't fix the causes. A more stable solution would be to use our new event bus to dispatch events when our resources are displayed and wait for this to do the scroll.

Well, I'd say let's try the more stable solution? :smile:

> I update the `text-blue-200` to a `\<strong\>` tag. We can use another color instead but this is not clickable.

I'm ok with the strong but a bit confused by the author text size set though. But nothing blocking.

theme/js/components/discussions/threads.vue
nicolaskempf57: This is a screenshot with bot the callout and the well to see the differences. I use green to match the previous design but I don't really know if it's necessary.
If we want a smaller callout, we can remove the button but the user will loose the ability to view all threads from here.
Capture d’écran 2021-12-01 à 11 52 54

nicolaskempf57: Callout only :
Capture d’écran 2021-12-02 à 11 01 38

Well only :

Capture d’écran 2021-12-02 à 11 02 28

nicolaskempf57: Our green well is more contrasted now.

maudetes (commented):
I'm not sure about the scroll event architecture. It looks a bit overkill if not needed elsewhere. What do you think? Do you think we would get use of it somewhere else?

Outside of this consideration, I think we're good? :)

nicolaskempf57:
> I'm not sure about the scroll event architecture. It looks a bit overkill if not needed elsewhere. What do you think? Do you think we would get use of it somewhere else?

I don't know of other use cases for this global event handling architecture. I don't think it's worth it to continue this for now, that's why I didn't go further to make a better solution for the scroll. I can go back to the timer version and we can reconsider it later when / if we decide to add this architecture.

maudetes (approved):
I think we're good ? :)

PR #42, opened 2021-11-26
Remove issues logic

quaxsze:


maudetes (commented):
And another 320 lines removed! :clap:!

I think there are some `macros/issues.html` imports left? Else it looks good to me!

PR #44, opened 2021-12-02
fix: add catch to avoid cancellation error

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/612

This PR also show a toast on search error instead of silently fail.


nicolaskempf57:
We can also move the cancel logic to an interceptor so it works for every request and not only the search one

maudetes (commented):
Thank you ! :)

I wonder if the `loading` shouldn't get set to `false` only when the request is a success (instead of finally)?

maudetes (approved):

PR #46, opened 2021-12-03
fix RGAA criterion Presentation 10.7

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/541


maudetes (approved):
Nice ! :D

PR #47, opened 2021-12-06
WIP : switch API suggest calls to APIV2

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/588


maudetes:
Will be done in https://github.com/etalab/udata-front/pull/60.

PR #50, opened 2021-12-07
Add topic in reuse search filter and meta information

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/638

Follows https://github.com/opendatateam/udata/pull/2689. It allows to display the topic information on reuse page in the meta-information tab if specified.
It also changes actual `type` filter in reuse search page by topic instead.


maudetes:
We can either keep the actual `tabs` or change them for DSFR or actual custom `tags`.

Alternative for tags using DSFR with custom active tag
image

Alternative for tags using actual tags with a new custom active tag
image

agarrone:
Exemple : https://www.gouvernement.fr/toute-l-actualite

theme/less/components/tabs.less
```diff
@@ -63,11 +63,10 @@ You can use tabs from DSFR. It uses DSFR colors and conventions. It allows you t
 
 .tabs {
     display: flex;
+    flex-wrap: wrap;
 
     .tab {
-        + .tab {
-            margin-left: @spacing-sm;
-        }
+        margin: @spacing-xxs;
```
nicolaskempf57, line 69: We can remove this line here and use DSFR utility class on reuse/list and home to allow for different margins (`.fr-m-1v` on reuse/list and `.fr-ml-3v` on home to keep the previous layout).

theme/less/components/tabs.less
```diff
@@ -63,11 +63,10 @@ You can use tabs from DSFR. It uses DSFR colors and conventions. It allows you t
...
 
         //Prevents children from taking the click events.
```
nicolaskempf57, line 71:
```suggestion
        // Prevents children from taking the click events.
```

theme/less/components/tabs.less
```diff
@@ -63,11 +63,10 @@ You can use tabs from DSFR. It uses DSFR colors and conventions. It allows you t
...
         //Might be dangerous, remove if needed
```
nicolaskempf57, line 72:
```suggestion
        // Might be dangerous, remove if needed
```

nicolaskempf57 (commented):
We can use DSFR utilities for margin/padding instead of our own spacing system so hopefully we'll be able to remove our custom utility classes in the future.

theme/less/components/tabs.less
maudetes: Will be fixed in https://github.com/etalab/udata-front/pull/57

theme/less/components/tabs.less
maudetes: Will be fixed in https://github.com/etalab/udata-front/pull/57

theme/less/components/tabs.less
maudetes: We'll use DSFR tags instead following https://github.com/etalab/udata-front/pull/57.

nicolaskempf57 (approved):

PR #52, opened 2021-12-13
Feat/update reuses style

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/577

This update reuses lists on display pages :
- 4 displayed
- with a pagination
- taller and thinner
- they have a fixed width instead of a width based on the viewport (not sure about this)

### DIFF
--- a/theme/js/components/reuse/card.vue
+++ b/theme/js/components/reuse/card.vue
@@ -2,7 +2,7 @@
 Vue and simpler -->
 
 <template>
-  <article class="reuse-card text-align-center fr-text--sm fr-mb-0">
+  <article class="reuse-card text-align-center fr-text--sm fr-mb-0" :class="{'reuse-card--fluid': fluid}">
     <Placeholder
       class="reuse-image fr-mb-3v fr-responsive-img"
       alt=""
@@ -38,7 +38,11 @@ export default {
     },
     private: {
       type: Boolean,
-      default: false
+      default: false,
+    },
+    fluid: {
+      type: Boolean,
+      default: false,
     }
   },
   components: {
--- a/theme/js/components/reuse/card.vue
+++ b/theme/js/components/reuse/card.vue
@@ -2,22 +2,22 @@
 Vue and simpler -->
 
 <template>
-  <article class="reuse-card">
+  <article class="reuse-card text-align-center fr-text--sm fr-mb-0">
     <Placeholder
-      class="reuse-image mb-sm"
-      :alt="title"
-      backgroundImage
+      class="reuse-image fr-mb-3v fr-responsive-img"
+      alt=""
       :src="image_url"
       type="reuse"
     />
-    <div class="reuse-info">
-      <div class="reuse-info-type fs-xs">
-        {{ type_label }}
-      </div>
-      <div class="reuse-info-amount fs-xs text-grey-300 mb-xs"></div>
-      <div class="reuse-info-title h4">
-        {{ title }}
-      </div>
+    <div>
+      {{ type_label }}
+    </div>
+    <div class="text-grey-300">
+      {{ $tc('X datasets', datasets) }}
+    </div>
+    <div>
+      {{ title }}
+      <span class="badge grey-300 fr-ml-1w" v-if="private">{{ $t('Private') }}</span>
     </div>
   </article>
 </template>
@@ -27,18 +27,27 @@ import Placeholder from "../utils/placeholder";
 import icon from "svg/actions/star.svg";
 
 export default {
-  created() {
-    this.icon = icon;
-  },
   props: {
     title: String,
     image_url: String,
     type_label: String,
     metrics: Object,
-    datasets: Number,
+    datasets: {
+      type: Number,
+      default: 0,
+    },
+    private: {
+      type: Boolean,
+      default: false
+    }
   },
   components: {
     Placeholder,
+  },
+  data() {
+   return {
+     icon
+   }
   },
 };
 </script>
--- a/theme/js/components/search/suggest-results.vue
+++ b/theme/js/components/search/suggest-results.vue
@@ -12,8 +12,8 @@ Used by the suggest feature to display typeahead-style results when you type you
 -->
 
 <template>
-  <div class="row p-md results">
-    <div class="p-md col col-md-12">
+  <div class="row fr-p-2w results">
+    <div class="fr-p-2w col col-md-12">
       <a
         class="unstyled row-inline justify-between see-all"
         :href="datasetUrl"
@@ -24,7 +24,7 @@ Used by the suggest feature to display typeahead-style results when you type you
         </h2>
         <span v-html="arrow" />
       </a>
-      <p class="text-grey-300 m-0">
+      <p class="text-grey-300 fr-m-0">
         {{
           $t(
             "Datasets are collections of data, i.e. structured information, easily readable by a machine."
@@ -40,7 +40,7 @@ Used by the suggest feature to display typeahead-style results when you type you
           :queryString="queryString"
           :link="datasetUrl"
         />
-        <div class="my-md cards-container reuse-cards" v-else>
+        <div class="fr-my-2w cards-container" v-else>
           <ul>
             <li v-for="dataset in results.datasets">
               <a :href="dataset.page" :title="dataset.title" class="unstyled">
@@ -48,13 +48,13 @@ Used by the suggest feature to display typeahead-style results when you type you
               </a>
             </li>
           </ul>
-          <a class="nav-link pt-md" :href="datasetUrl">{{
+          <a class="nav-link fr-pt-2w" :href="datasetUrl">{{
             $t("Search in datasets")
           }}</a>
         </div>
       </transition>
     </div>
-    <div class="p-md col col-md-12">
+    <div class="fr-p-2w col col-md-12">
       <a
         class="unstyled row-inline justify-between see-all"
         :href="reuseUrl"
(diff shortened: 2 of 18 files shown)


nicolaskempf57:
Here are some details about the fixed width vs viewport based width : https://www.figma.com/exit?url=https%3A%2F%2Fconciergerie.notion.site%2FR-utilisations-largeur-variable-ou-largeur-fixe-510c6ff46ec14996ac272583b016e8bf&fuid=1024706569346348736

nicolaskempf57:
We can add an hover effect to reuses like the DSFR link hover below
Capture d’écran 2022-01-03 à 09 23 22

maudetes (commented):
Thank you for this PR!

I have some comments:
- on homepage with mobile, we can't scroll all the way to the right, the last reuse is cut. Example with 400*750 device
image
- _twice_, on a reuse horizontal list, one of the reuse was half the width of the others. But I haven't reproduced it yet, I'll try to! :sweat_smile:
- could we add a test on reuse pagination on dataset page?

Additionally, I noticed that [reuse pagination on datasets page](https://dev.data.gouv.fr/fr/datasets/demandes-de-valeurs-foncieres/?reuses_page=2#community-reuses) doesn't work on dev (with merged demo branch https://github.com/etalab/udata-front/pull/58) but works locally. I don't think we need to debug this, but we should pay attention when merging later.

udata_front/theme/gouvfr/templates/reuse/card.html
```diff
@@ -1,31 +1,26 @@
...
-                        {{ reuse.type_label }}
-                    </span>
-                    <div class="reuse-info-amount fs-xs text-grey-300 mb-xs">
-                        {{ reuse.datasets|length or 0 }} {{ _('datasets') }}
-                    </div>
-                </div>
-                <div class="relative">
-                    <span class="text-orange-100">{% include theme('svg/actions/star.svg') %}</span>
-                    <span class="badge orange-100 ml-xxs py-xxxs px-xs">{{ reuse.metrics.followers or 0 }}</span>
-                    <span class="visually-hidden">{{ _('favourites') }}</span>
```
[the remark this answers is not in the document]
nicolaskempf57: It's not on Figma so I removed it but I don't know if it's a mistake or not 🤔 I don't remember that we talk about it. I'll see with @rkalfane

udata_front/theme/gouvfr/templates/home.html
```diff
@@ -149,13 +149,13 @@ <h2>{{ _('Data reuses') }}</h2>
             </div>
         </div>
     </div>
-    <div class="hagrid reuse-cards">
+    <div class="hagrid">
         {% for tab_id, label, reuses, kwargs in reuses_tabs  %}
-        <div class="tab-content  {% if loop.first %}active{% endif %}" id="{{tab_id}}">
-            <div class="row pt-xl pt-md-md">
+        <div class="tab-content {% if loop.first %}active{% endif %}" id="{{tab_id}}">
+            <div class="row pt-xl pt-md-md overflow-x-scroll no-wrap">
```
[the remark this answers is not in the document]
nicolaskempf57: Thanks ! This was not showing because of my OS config.

nicolaskempf57:
> * on homepage with mobile, we can't scroll all the way to the right, the last reuse is cut. Example with 400*750 device

The scroll is fixed on small devices

> * _twice_, on a reuse horizontal list, one of the reuse was half the width of the others. But I haven't reproduced it yet, I'll try to! 😅

I'm also trying to reproduce. For now, the last time I thought I find a reproduction, it was actually me forgetting to do an `npm install` after going back to my branch.

nicolaskempf57:
> * could we add a test on reuse pagination on dataset page?

Done !

nicolaskempf57:
After this morning design meeting, we now use col-based width for reuses with a max-width to avoid other aspect ratio. We'll be able to use aspect-ratio CSS property when we remove older browser support.

The only opened issue on this PR is https://github.com/etalab/udata-front/pull/52/files#r779027223 regarding the number of datasets on suggest (not present in api call).

udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -397,10 +397,9 @@ <h2 id="community-reuses" class="fr-h2 fr-col">{{ _('Reuses') }} <sup>{{ total_r
         </div>
         <p class="fr-my-0">{{ _('Explore the reuses of this dataset.') }}</p>
         <p class="fr-mt-0 fr-mb-3w">{{ _('Did you use this data ? Reference your work and increase your visibility.') }}</p>
-        <div class="fr-pt-2w fr-pt-md-0 fr-grid-row no-wrap fr-mb-2w overflow-x-scroll">
```
maudetes: Do we want to say bye bye `overflow-x-scroll` or could it get useful later?

theme/less/specific/reuse.less
```diff
@@ -1,14 +1,12 @@
...
     .reuse-image {
         border-radius: @spacing-md;
         height: 480px;
         object-fit: cover;
         object-position: center center;
-    }
-
-    &.reuse-card--fluid {
-        width: 100%;
+        // aspect-ratio: 3 / 5; // We can use an aspect-ratio instead of a fixed height when
```
maudetes, line 10: when ... ? (suspense :p)

udata_front/theme/gouvfr/templates/dataset/display.html
nicolaskempf57: I removed it. We can add it back later if needed.

theme/js/components/reuse/card.vue
```diff
@@ -2,7 +2,7 @@
 Vue and simpler -->
 
 <template>
-  <article class="reuse-card text-align-center fr-text--sm fr-mb-0" :class="{'reuse-card--fluid': fluid}">
```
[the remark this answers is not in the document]
nicolaskempf57: Yes !

theme/less/specific/reuse.less
nicolaskempf57: We can use an aspect-ratio instead of a fixed height when we update our list of supported browsers.
This will not work on IE 11 and is only working on 82% of browsers on 01/2022.
Updated :)

theme/js/components/reuse/card.vue
```diff
@@ -2,22 +2,22 @@
...
-      </div>
-      <div class="reuse-info-amount fs-xs text-grey-300 mb-xs"></div>
-      <div class="reuse-info-title h4">
-        {{ title }}
-      </div>
+    <div>
+      {{ type_label }}
+    </div>
+    <div class="text-grey-300">
+      {{ $tc('X datasets', datasets) }}
```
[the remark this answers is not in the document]
maudetes: Indeed, I guess we can remove this information from the vue card for now?

theme/js/components/search/suggest-results.vue
```diff
@@ -65,31 +65,31 @@ Used by the suggest feature to display typeahead-style results when you type you
...
           :link="reuseUrl"
         />
-        <div class="my-md cards-container" v-else>
+        <div class="fr-my-2w cards-container" v-else>
           <ul class="row">
-            <li v-for="reuse in results.reuses" class="col text-align-center">
+            <li v-for="reuse in results.reuses" class="col-6">
               <a :href="reuse.page" :title="reuse.title" class="unstyled">
-                <Reuse v-bind="reuse" />
+                <Reuse v-bind="reuse" :fluid="true" />
```
maudetes, line 88: I don't think we use the `fluid` prop anymore?

udata_front/theme/gouvfr/templates/reuse/card.html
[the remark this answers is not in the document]
maudetes: When last discussed with @agarrone, this topic was still opened, wasn't it? Should we keep favourites on reuse card or not? :)

theme/js/components/reuse/card.vue
```diff
@@ -2,43 +2,46 @@
 Vue and simpler -->
 
 <template>
-  <article class="reuse-card">
+  <article class="reuse-card text-align-center fr-text--sm fr-mb-0">
     <Placeholder
-      class="reuse-image mb-sm"
-      :alt="title"
-      backgroundImage
+      class="reuse-image fr-mb-3v fr-responsive-img"
```
maudetes, line 7: I think the ratio on suggest can still be squeezed like this (on a 900*750 screen):
image

maudetes (commented):
I think we're almost good on this one!
Some minor comments left and a question on design final decision on favourites :)

agarrone:
I've checked : we did decided not to show favorites here. At least as long as it gives no informations to users 🙂

theme/js/components/reuse/card.vue
nicolaskempf57: I removed some padding on suggest to improve this in f288583.

maudetes (approved):
Don't forget the changelog entry.

I think we're almost good on this PR, thank you! :pray:

theme/js/locales/en.json
```diff
@@ -118,5 +118,7 @@
   "Copy permalink": "Copy permalink",
   "Discussions": "Discussions",
   "Error getting search results.": "Error getting search results.",
-  "Sort discussions by": "Sort discussions by"
+  "Sort discussions by": "Sort discussions by",
+  "X datasets": "0 datasets | 1 dataset | {n} datasets",
```
maudetes, line 122: We don't need `X datasets` translation anymore for now?

PR #54, opened 2021-12-14
Change wording in participez

agarrone:
See https://github.com/etalab/data.gouv.fr/projects/4#card-64278487


maudetes (commented):
I think you should do these changes in [crowdin](https://crowdin.com/project/udata-front) instead if you just want to change the translation :)

PR #57, opened 2022-01-04
Use DSFR tag instead of .btn-tab class

nicolaskempf57:


theme/less/components/tabs.less
```diff
@@ -9,26 +9,17 @@ Tabs allow for easy navigation between categories. You can toggle a `.active` cl
 Warning : `.tab` first immediate child (`.tab > *`) has a `pointer-events: none` property set so that clicks actually register to the tab element and not the children.
 This prevents messy event handling with bubbling prevention, but might cause you trouble in some edge cases.
 
-```btn-tabs.html
-<nav class="tabs">
-    <a class="btn-tab tab active" href="#">Selection Etalab</a>
-    <a class="btn-tab tab" href="#">Dernières réutilisations</a>
-    <a class="btn-tab tab" href="#">Voir plus</div>
+```tabs.html
```
maudetes, line 12: Do you think that we should keep  these `tabs` and `tab` in the future? Or could we remove those and use dsfr tabs or tags and move the logic left in tags if needed?

theme/less/components/tags.less
```diff
@@ -15,8 +15,30 @@ They also come in a bigger size with the `.big` class.
...
 <span class="big tag">Big tag</span>
 ```
+
+## DSFR tags
+
+We can use DSFR tags and their customizations. See Tabs-js for how to use it as tabs.
+There is also `.fr-tags-group` to group multiple tags together like a row.
+
+```dsfr-tags.html
+<nav class="tabs fr-tags-group">
```
maudetes, line 25: I think we can remove `tabs` here?

maudetes (commented):
Thank you a lot for this PR ! :clap:

I have some question on the future of `tabs`, but else it looks good to me.

theme/less/components/tabs.less
nicolaskempf57: We should remove this custom logic and rely on DSFR tabs only in the future to avoid usability and accessibility differences.

But in the meantime, I think it's best to keep them separate so we don't change the default behavior  of tags. Our current tab logic breaks the anchor typical flow.

theme/less/components/tags.less
nicolaskempf57: Yes ! Tabs and tab have nothing to do with tags style

theme/less/components/tabs.less
```diff
@@ -61,36 +52,15 @@ You can use tabs from DSFR. It uses DSFR colors and conventions. It allows you t
...
-        color: white; //Defaults to white for legibility
-        padding: @spacing-xs @spacing-md;
-
-        transition: opacity ease 0.2s;
-
-        &:hover, &.active {
-            opacity: 1;
-        }
+    &.active {
+        pointer-events: none;
```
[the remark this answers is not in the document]
nicolaskempf57: I added back the `.tabs` requirement for this rule. This prevents any conflict with our header.

maudetes (approved):
I think we're all good, thank you !

PR #60, opened 2022-01-06
Use refactored search adapter

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/588

Following https://github.com/opendatateam/udata/pull/2692, we use the refactored search adapter.

Default lists page without query (`/fr/datasets/`) has the same behavior as previously.

We are merging on the new `use-search-v2` branch and not directly on master.


theme/js/components/search/suggestor.vue
```diff
@@ -83,6 +84,7 @@ export default {
   },
   props: {
     onChange: Function,
+    suggestApi: Function,
```
nicolaskempf57, line 87: Does an Object with a default value work instead of the check l.141. ?

I don't know if `this` is available here, otherwise we can export api from `js/plugins/api`. 
```suggestion
    suggestApi: {
        type: Object,
        default: () => this.$api,
    },
```

nicolaskempf57 (commented):
I just looked at the files and didn't test it on local yet.

theme/js/components/search/suggestor.vue
maudetes: Thank you for this input, indeed it would be nicer to have a default value here.
However, we're switching back to apiv1 for suggest endpoints (to prevent breaking changes), so we won't need the `suggestApi` prop anymore.

theme/js/components/dataset/card.vue
```diff
@@ -16,14 +20,31 @@
...
 import Placeholder from "../utils/placeholder";
+import certified from "svg/certified.svg";
+import privateIcon from "svg/private.svg";
+import useOrganizationCertified from "../../composables/useOrganizationCertified";
+import useUserAvatar from "../../composables/useUserAvatar";
 
 export default {
-  components: {
-    Placeholder,
+  created() {
```
nicolaskempf57, line 29: Do you want to refactor this component to use the new Composition API ?

theme/js/components/reuse/card.vue
```diff
@@ -6,11 +6,11 @@ Vue and simpler -->
...
       class="reuse-image fr-mb-3v fr-responsive-img"
       alt=""
-      :src="image_url"
+      :src="image"
       type="reuse"
     />
-    <div>
-      {{ type_label }}
+    <div class="text-grey-300">
+          {{ $tc('X datasets', datasets.length) }}
```
nicolaskempf57, line 13: Does the new API return a list of datasets or only a number ? The props says `datasets` is a Number but it's used as an Array here.

maudetes:
Changes on suggest will be obsolete after https://github.com/etalab/udata-front/pull/88.

nicolaskempf57 (approved):

PR #63, opened 2022-01-11
Add temporal coverage back

nicolaskempf57:
Regarding https://github.com/etalab/data.gouv.fr/issues/547, it seems that the file is not parsed by `inv i18n`.

We can also update the code to use `format_date()` short format instead of making our owns.

Close https://github.com/etalab/data.gouv.fr/issues/547


maudetes (approved):
Nice !

Indeed for translation, we can go back to use `[python: **.py]` in babel.cfg.

nicolaskempf57:
I added `[python: **/frontend/**.py]` in babel.cfg. Is it best to use `[python: **.py]` ?

maudetes:
> I added `[python: **/frontend/**.py]` in babel.cfg. Is it best to use `[python: **.py]` ?

I think that we would miss some translations as well, ex in the folder: `udata_front/views/`

PR #64, opened 2022-01-12
Don't wrap thread header above md screens

nicolaskempf57:
Header and copy buttons aren't wrapped to a new line on lg screen and above. They are wrapped for md screens and below. Here are some screenshots to see the wrap / nowrap for :
1. desktop
2. lg screen just above md
3. md screen

Capture d’écran 2022-01-12 à 10 00 34

Capture d’écran 2022-01-12 à 10 01 11

Capture d’écran 2022-01-12 à 10 01 21

We can add more changes if we want to :

- The current text-wrap is the browser default, we can update it if we want to keep the link on a single line for example.
- There is no col with width in this PR (like col-8), we can use this solution if we want more control about the ratio between title and link.
- 
Close https://github.com/etalab/data.gouv.fr/issues/670


maudetes (approved):
I think we're all good :)

PR #65, opened 2022-01-12
Fix read-more height error with img

nicolaskempf57:
Vue.js has no way to know if img are loaded and in markdown description, imgs don't have any height from CSS so their sizes is based on the file size.

To solve this error, we redo the calculation after some time. We don't have any way to easily know ourself if imgs are loaded or not so it's a flat 500ms.

This seems to fix https://github.com/etalab/data.gouv.fr/issues/667

Close https://github.com/etalab/data.gouv.fr/issues/669


theme/js/components/utils/read-more.vue
```diff
@@ -72,6 +67,16 @@ export default {
               ease: easing.anticipate,
             }).start(divStyler.set);
           }
+    },
+    updateReadMoreHeight() {
+      let contentHeight = Array.from(this.$refs.container.children)
+      .map(getHeight)
+      .reduce((total, height) => total + height, 0)
+      this.readMoreRequired = contentHeight > getHeight(this.$refs.container);
+      this.containerHeight = DEFAULT_HEIGHT;
```
maudetes, line 76: Why do we need to set `containerHeight` again?

maudetes (commented):
It seemed to have fixed both issues, even if difficult to reproduce!

theme/js/components/utils/read-more.vue
nicolaskempf57: Explained :)

maudetes (approved):
Seems all good :)

PR #67, opened 2022-01-13
Check if datasets and reuses are visibles before showing them

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/613

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Fix hidden datasets shown on Home and Reuses [#67](https://github.com/etalab/udata-front/pull/67)
 
 ## 1.2.1 (2022-01-11)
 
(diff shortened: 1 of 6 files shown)


maudetes (commented):
I think that maybe instead of filtering on is_visible, we could also filter on private.
Indeed, a dataset without resources or archived wouldn't be visible: https://github.com/opendatateam/udata/blob/a295cab3c0e8f086fea1853655011f361ac81b77/udata/core/dataset/models.py#L505.

However, we have used visible at some other places in the front (on organization page for datasets and reuses lists) and we could keep this for coherence.

What do you think?

nicolaskempf57:
> However, we have used visible at some other places in the front (on organization page for datasets and reuses lists) and we could keep this for coherence.
> 
> What do you think?

I chose `is_visible` to stay consistent with other usages on home (recent datasets and reuses) and organization page. Also the name `visible` seems like it was made for this use case.

maudetes (approved):
Nice ! :)

I think you can update or rollback the translations though (see https://github.com/etalab/udata-front/pull/63#issuecomment-1015474428)

PR #68, opened 2022-01-13
Add query param message to messages displayed in well

nicolaskempf57:
This PR is based on the current flash system (https://github.com/gmougeolle/udata-gouvfr/commit/a6856663c9884171e3d1d0fb3bc1fab399d37f6f) and the current $auth implementation (https://github.com/gmougeolle/udata-gouvfr/commit/6ee183b0106ec433e803207fba1c98959eeeecbb)

It only adds the missing glue to show the message sent from $auth.

We may want to restrict the display of query parameter message to the login page to avoid abuses.

Close https://github.com/etalab/data.gouv.fr/issues/316


maudetes (commented):
This seems to do the work, but we must pay attention not to use the `message` param anywhere?

Additionally, did you find usage of `flashed` messages somewhere? I couldn't find anything (`flash()` never called, no mention in flask restplus either) but maybe I'm missing something?

nicolaskempf57:
We must pay attention to this and anybody can make link with any message like `http://dev.local:7000/en/datasets/?message=Rendez-vous%20sur%20www.somesite.fr%20pour%20plus%20d%27informations`.

Flashed messages are used by the authorization flow for actions requiring a logged user like the community resource example in the issue  https://github.com/etalab/data.gouv.fr/issues/255

maudetes:
Could we use generic flashed message if needed instead (as in actual flashed messages: ) ? This way we could just add a param `login_required=true`, and we would prevent users to fiddle with the message.

nicolaskempf57:
This would be a safer option to do. The only drawback is that we'll loose the custom message that say why the login is required.

maudetes:
Let's go for the safer option with a generic message for now! :pray:

nicolaskempf57:
Done !

maudetes (approved):
I think we're all clear!

PR #72, opened 2022-01-25
Show search result hover on focus

nicolaskempf57:
I updated CSS rules to allow `<a>` focus to show hover data and background. To be consistent, I moved the `:hover` to the `<a>` too.

In the future, we can refactor `dataset/search-result.vue` and `dataset/search-result.html` to include the <a> instead of duplicating it for each parent file.

Close https://github.com/etalab/data.gouv.fr/issues/542


maudetes (approved):
Works like a charm :sparkles:

PR #73, opened 2022-01-25
Add banner for harvested dataset

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/657


maudetes (commented):
The banner seems good to me.

However, I think that `remote_url` is actually only set for CKAN datasets. I'm not sure we can have one consistent value. I need to check that up. At worst, we can use `harvest:domain`.

maudetes:
> However, I think that `remote_url` is actually only set for CKAN datasets. I'm not sure we can have one consistent value. I need to check that up. At worst, we can use `harvest:domain`.

It is set for CKAN and ODS datasets but not for DCAT. But we should keep on using `remote_url` here and make sure it is set in DCAT in the backend.

maudetes (approved):
I think it's all good here ! :tractor:

maudetes (changes requested):
After discussion, we need to change the background color, that seems like a warning. We will also change for a simple background and grid instead of the well for a better icon alignment.

nicolaskempf57:
Banner background is now blue !

udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -101,6 +102,14 @@ <h1 class="fr-mb-0 fr-h1">
             {% endif %}
         </div>
     </div>
+    {% if external_source %}
+        {{ banner_info(
+            "fr-fi-external-link-line",
+            _("This dataset come from an external portal.
```
maudetes, line 108: I think we're missing a space between the two phrases here
image

maudetes (approved):

PR #75, opened 2022-02-01
Remove underline from button hover

nicolaskempf57:
This PR solve the inconsistency regarding button hover. There were 3 cases of different hovers :
- underlined text from our UI Kit
- colored background from DSFR (not used alone in udata-front)
- underlined text with colored background from buttons with both classes from the UI Kit and DSFR

To do so, I merge our button classes with DSFR ones. 
The only changes left are our own color, the border-radius and an override due to our `a:hover` rule.

Sometimes, it was easier to change our custom made component to a DSFR one instead of merging some styles. So this PR also remove our dropdown and tags to use DSFR ones.

Close https://github.com/etalab/data.gouv.fr/issues/668
Close https://github.com/etalab/data.gouv.fr/issues/689
Close https://github.com/etalab/data.gouv.fr/issues/692


theme/less/components/buttons.less
```diff
@@ -132,87 +98,29 @@ There are `.fr-btn--icon-left` and `fr-btn--icon-right` to add spacing around th
     }
 }
 
-
-.btn-primary {
-    .btn;
```
maudetes, line 137: `font-size` is 1rem now, instead of `@font-size-sm` with previous buttons (see `.btn` class). Can we get back to previous font size?

maudetes (commented):
Thank you for this PR!

Some minor comments :)

theme/js/components/dataset/preview.vue
```diff
@@ -5,7 +5,7 @@
     </div>
 
     <footer class="modal-footer">
-      <button class="btn-primary" @click.prevent="close()"> X </button>
+      <button class="fr-btn fr-fi-close-line" @click.prevent="close()"> {{$t('Close')}} </button>
```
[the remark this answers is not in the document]
nicolaskempf57: Yes ! We can add a title attribute too for every user. The DSDR documentation use both at the same time.

theme/less/components/dropdown.less
```diff
@@ -1,84 +0,0 @@
...
-<div class="dropdown">
-    <select>
-        <option>Option A</option>
-        <option>Option B</option>
-    </select>
-</div>
-```
-*/
-
-.dropdown {
```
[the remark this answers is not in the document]
nicolaskempf57: We can remove the `search.html` file but we'll have to remove the entire topic directory because `search.html` is used inside `subnav.html` and this one is included in the `display_base.html` extended in all other files.

theme/less/components/buttons.less
```diff
@@ -5,67 +5,49 @@ category: Components
...
 
 ## Primary buttons
-The `.btn` classes are designed to be used with the `<button>` element.
-
+The `.fr-btn` classes are designed to be used with the `<button>` element.
 However, you can also use these classes on `<a>` or `<input>` elements (though some browsers may apply a slightly different rendering).
-When using button classes on `<a>` elements that are used to trigger in-page functionality (like collapsing content), rather than linking to new pages or sections within the current page, these links should be given a `role="button"` to appropriately convey their purpose to assistive technologies such as screen readers.
 
-Primary buttons exist for every color in this UI-kit. Please refer to the [colors page](/#colors) to see available text and background colors.
+When using button classes on elements that are used to trigger in-page functionality (like collapsing content), use a <button>.
```
maudetes, line 11:
```suggestion
When using button classes on elements that are used to trigger in-page functionality (like collapsing content), use a `<button>`.
```

theme/less/components/buttons.less
```diff
@@ -5,67 +5,49 @@ category: Components
...
+<p><a href="#" role="button" class="fr-btn fr-btn--secondary fr-btn--secondary-green-400">fr-btn--secondary-green-400</a></p>
 ```
 
-## Disabled state
-Make buttons look inactive by adding the disabled boolean attribute to any `<button>` element.
-
-`<a>`s don’t support the disabled attribute, so you must add the `.disabled` class to make it visually appear disabled.
-Disabled buttons should include the `aria-disabled="true"` attribute to indicate the state of the element to assistive technologies.
+## Buttons with
+icons
```
maudetes, line 40:
```suggestion
## Buttons with icons
```

maudetes (approved):
I think we're good :)

PR #88, opened 2022-03-07
Replace suggest

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/747


theme/less/components/well.less
```diff
@@ -61,6 +61,10 @@ They exist in many colors and two versions, primary with a solid background and
         vertical-align: baseline;
     }
 
+    [class^="fr-fi-"] {
```
maudetes, line 64: I think this rule is already defined above.

udata_front/theme/gouvfr/templates/header.html
```diff
@@ -13,110 +15,164 @@
...
+                                        class="fr-link fr-fi-logout-box-r-line"
+                                    >
+                                        {{ _('Logout') }}
+                                    </a>
+                                </li>
+                            {% else %}
+                                <a
+                                    href="{{ url_for('security.login', next=next_url) }}"
+                                    title="{{ _('Log in') }}"
+                                    class="fr-link fr-fi-lock-line"
```
maudetes, line 104: Icons are underlined when hovered

maudetes (commented):
I'm still amazed by this new suggest that isn't a suggest! :star_struck: 

Another batch of review ;)

udata_front/theme/gouvfr/templates/header.html
nicolaskempf57: I removed the underline hover for DSFR link because they already have an hover state. This match what we did with buttons.

theme/js/components/search/menu-search-option.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <a class="fr-nav__link unstyled" :href="link" target="_self">
+    <div class="fr-grid-row fr-grid-row--middle">
+      <div aria-hidden="true" class="fr-col-auto fr-mr-2w" v-html="icon">
+      </div>
+      <i18n-t keypath="Search X in Y" class="fr-col" tag="div" v-if="query">
+        <template #query>
+          &laquo;<em>{{query}}</em>&raquo;
```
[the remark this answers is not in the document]
nicolaskempf57: It seems that the suggested way to handle this is to add them to the string to translate.

udata_front/theme/gouvfr/templates/header.html
```diff
@@ -13,110 +15,164 @@
...
+                <div class="fr-header__brand fr-enlarge-link">
+                    <div class="fr-header__brand-top">
+                        <div class="fr-header__logo">
+                            <p class="fr-logo">
+                                République
+                                <br>Française
+                            </p>
+                        </div>
+                        <div class="fr-header__navbar">
+                            <button
```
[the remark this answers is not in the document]
nicolaskempf57: This seems to work fine on Firefox & Chrome based browsers when using native "disable javascript" function but not with ublock or other ad blocker that erase `<noscript>` tag and its classes.

theme/js/composables/useBreakpoints.js
```diff
@@ -0,0 +1,32 @@
...
+ * @property {BreakpointModel} XL
+ */
+const Breakpoints = window.dsfr.core.Breakpoints;
+/**
+ * @function isBreakpoint
+ * @param {BreakpointModel | string} breakpoint
+ * @returns boolean
+ */
+const isBreakpoint = window.dsfr.core.Instance.prototype.isBreakpoint;
+export default function useBreakpoints() {
```
maudetes, line 24: I don't think it is used, is it?

maudetes (commented):
On keyboard navigation, when focus is lost, the dropdown doesn't collapse.

theme/js/composables/useActiveDescendant.js
```diff
@@ -0,0 +1,113 @@
...
+   */
+  const handleKeyPressForActiveDescendant = (key, alreadyMovedDown = false) => {
+    switch (key.keyCode) {
+      case KEYCODES.DOWN:
+        if(!alreadyMovedDown && !key.altKey) {
+          selectNextOption();
+        }
+        break;
+      case KEYCODES.UP:
+        selectPreviousOption();
```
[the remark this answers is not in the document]
nicolaskempf57: We don't want the optional "Alt + key up" interaction as defined in the specification. It closes the popup and moves the visual focus to the input field. This is great for combobox in forms but not for our menu use case.

nicolaskempf57:
> I think we have an issue on mobile. When clicking (instead of using `Enter`) on one of the options, it doesn't follow the link but collapse the options instead.

I reproduce it also on desktop sometimes. This was a conflict between our input blur event and the link click.

theme/js/components/search/menu-search.vue
```diff
@@ -0,0 +1,174 @@
...
+        >
+          <MenuSearchOption :icon="option.icon" :query="q" :type="option.type" :link="option.link"/>
+        </li>
+      </ul>
+    </div>
+  </div>
+</template>
+
+<script>
+import {ref, reactive, onMounted, onUnmounted, nextTick} from "vue";
```
maudetes, line 73:
```suggestion
import {ref, reactive, onMounted, onUnmounted} from "vue";
```

maudetes (approved):
I think it's all clear, well done for this! :clap:

PR #91, opened 2022-03-07
Replace see more button with link

nicolaskempf57:
There is now a default gutter between tags and the "see more" link. We can remove it if we want to.

Capture d’écran 2022-03-07 à 13 27 01

Fix https://github.com/etalab/data.gouv.fr/issues/262

### DIFF
--- a/theme/less/content/links.less
+++ b/theme/less/content/links.less
@@ -133,3 +133,10 @@ a {
 .fr-link.rounded-0 {
     .rounded-0;
 }
+
+.fr-link {
+    --text-action-high-blue-france: @blue-400;
+    &:hover {
+        text-decoration: none;
+    }
+}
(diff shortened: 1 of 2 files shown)


maudetes (commented):
I find it weird that the button `see more` appears before the featured objects on mobile, instead of after?

Additionally, I find it a bit weird to have a different font size between the tags and the link, what do you think?

nicolaskempf57:
> I find it weird that the button `see more` appears before the featured objects on mobile, instead of after?

We can move the button after the lists like in organization page. Here is an example with the same button style :

Capture d’écran 2022-03-22 à 09 05 59

agarrone:
I kind of prefer the first version, but the second one is ore coherent with the behaviour we expect on the organisation page. 
Most of the time we use the top right corner for action buttons

maudetes:
I suggest we keep the first version then, we can iterate on these later.

PR #96, opened 2022-03-24
Feat/add static html pages

maudetes:
Fix https://github.com/etalab/data.gouv.fr/issues/776

Add more dsfr components that will be used by HTML static pages:
- callout
- card
- highlight
- quote
- sidemenu
- tile

You can test HTML pages rendering with http://dev.local:7000/fr/pages/test/test/.
This renders https://github.com/etalab/datagouvfr-pages/blob/master/pages/test/test.md.

On a first iteration, I used the same logic and template for HTML pages as for Markdown.
A `content_type` can be set to `html` (default to `markdown`) using frontmatter.
In case of html, the [safe](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.safe) filter is used instead of `markdown`.

We could also
- discriminate on the path, having `pages/markdown/...` and `pages/html/...`, etc.
- discriminate on the file extension (not specified in slug today, expecting `.md` only)
- have a different repo for HTML entirely
- ... ?


nicolaskempf57 (approved):

maudetes:
@abulte interested on your point of view to deal with both md and html static files :)

maudetes:
> I would discriminate on extension, it'll probably be easier to understand.

Did you think of something like: https://github.com/etalab/udata-front/pull/96/commits/4f2f9b01e60ceb6869b7b0f066843a853e29f15a? It requires an additional HEAD request though.

abulte:
@maudetes more complicated than I thought, sorry :p (I forgot the extension was not in the URL). But quite elegant! 🤝

udata_front/views/gouvfr.py
```diff
@@ -54,11 +54,18 @@ def get_pages_gh_urls(slug):
...
-    raw_url = f'https://raw.githubusercontent.com/{repo}/{branch}/pages/{slug}.md'
-    gh_url = f'https://github.com/{repo}/blob/{branch}/pages/{slug}.md'
+    raw_url = f'https://raw.githubusercontent.com/{repo}/{branch}/pages/{slug}'
+    gh_url = f'https://github.com/{repo}/blob/{branch}/pages/{slug}'
+
     return raw_url, gh_url
 
 
+def detect_pages_extension(raw_url):
+    if requests.head(raw_url + '.md').status_code == 200:
```
abulte, line 64:
```suggestion
    if requests.head(f'{raw_url}.md').status_code == 200:
```

udata_front/views/gouvfr.py
```diff
@@ -101,15 +113,15 @@ def get_object(model, id_or_slug):
 
 @blueprint.route('/pages/<path:slug>/')
 def show_page(slug):
-    content, gh_url = get_page_content(slug)
+    content, gh_url, content_type = get_page_content(slug)
```
abulte, line 116: Maybe keep `extension` here to stay coherent?

abulte (approved):

PR #102, opened 2022-04-13
Add information about required fields

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/543


maudetes (commented):
Thank you for this fix!

Could we improve the layout to make it less confusing with the other fields?

Ex: aligning it right and matching the star color:
image

nicolaskempf57:
> Could we improve the layout to make it less confusing with the other fields?

I don't really like the right alignment but the color matching is nice !
We can use a sentence like the recommandation from RGAA 3. Do you prefer this one ?

Capture d’écran 2022-04-25 à 15 19 17

maudetes:
> I don't really like the right alignment but the color matching is nice ! We can use a sentence like the recommandation from RGAA 3. Do you prefer this one ?

Yes, I like this suggestion better :)

theme/js/components/discussions/thread-reply.vue
```diff
@@ -9,7 +9,15 @@
       </div>
     </div>
     <form class="form fr-p-0" @submit.prevent="submit">
-      <p class="fr-mt-0 fr-mb-1w fr-text--sm">{{$t('* : Required field')}}</p>
+      <i18n-t
+        keypath="Fields preceded by a star ({markup}) are required."
```
maudetes, line 13: I think we should use `followed` instead of `preceded` everywhere?

theme/js/components/discussions/thread-reply.vue
```diff
@@ -9,7 +9,15 @@
...
+        tag="p"
+        class="fr-mt-0 fr-mb-1w fr-text--sm"
```
maudetes, line 15: Can I suggest having a smaller grey text ? Ex: `fr-text--xs text-grey-380`?

maudetes (commented):
Nice, thank you!

theme/js/components/discussions/thread-reply.vue
nicolaskempf57: I'll try this !

theme/js/components/discussions/thread-reply.vue
maudetes: aaaah indeed ^^

maudetes: let's keep the recommended sentence :)

theme/js/components/discussions/thread-reply.vue
nicolaskempf57: I only put `fr-text--xs` for now because I found it hard to read with both. Would it work for you ?

maudetes (approved):

PR #104, opened 2022-04-14
feat: add label for logo and hide decorative svg

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/805


maudetes (approved):
I think we're good! :clap:

PR #108, opened 2022-05-10
Add lighthouse CI

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/483


.circleci/config.yml
```diff
@@ -148,6 +151,35 @@ jobs:
...
+      - image: redis:alpine
+    steps:
+      - checkout
+      - browser-tools/install-chrome
+      - attach_workspace:
+          at: .
+      - run:
+          name: Generate udata fixtures
+          command: |
+            pip install -e .
```
maudetes, line 168: We reinstall package because we can't use venv since we're on a next-gen CircleCI image using pyenv.

lighthouserc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+    ci: {
+      collect: {
+        startServerCommand: 'npm run serve',
+        url: ['http://localhost:7000/en/']
+      },
+      upload: {
```
maudetes, line 7: We could have different uploads solution. See https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#upload

Here we're storing export on filesystem and saving it as artifact in CI.

.circleci/config.yml
```diff
@@ -148,6 +151,35 @@ jobs:
...
+            udata generate-fixtures
+      - run:
+          name: Install lighthouse
+          command: sudo npm install -g @lhci/cli@0.8.x
```
maudetes, line 172: We could also add it in devDependencies in package.json directly

maudetes (commented):
Adding lighthouse in CircleCI.

For now, it never fails, we could add assertions to fail if some conditions aren't met. See https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#assert.

This additional job does not make the pipeline longer since it runs in parallel to `python` job mostly.

We could probably keep most of the logic for Cypress if we want to add Cypress tests: https://docs.cypress.io/guides/continuous-integration/circleci (related to https://github.com/etalab/data.gouv.fr/issues/781).

quaxsze:
👍

maudetes:
I added assertions rules based on current results score to prevent regressions.
We should however fix most of the issues, in order to use the recommended assertion preset (`"preset": "lighthouse:no-pwa"`). It is now currently failing on too many errors.

This should be done in a different issue: https://github.com/etalab/data.gouv.fr/issues/894.

maudetes (approved):

PR #113, opened 2022-06-07
Add new menu items

nicolaskempf57:
This PR add new menu items as nested items.

Nested items are shown in a collapsible menu. 
Ids and aria-* are used to make it accessible.

Previous jinja call to check if an item is the current endpoint is now an helper called `is_current_endpoint`.
There is a new jinja helper to check a match for both the endpoint and the corresponding args called `is_current_page`.
The new one is used for nested items.

Close https://github.com/etalab/data.gouv.fr/issues/846
Close https://github.com/etalab/data.gouv.fr/issues/479

### DIFF
--- a/udata_front/frontend/menu_helpers.py
+++ b/udata_front/frontend/menu_helpers.py
@@ -24,5 +24,4 @@ def is_current_page(request: Request, item: Item) -> bool:
 
 @front.app_template_global()
 def is_parent_of_current_endpoint(request: Request, items: List[Item]) -> bool:
-    current_item_in_items = map(lambda item: is_current_page(request, item), items)
-    return any(current_item_in_items)
+    return any(is_current_page(request, item) for item in items)
--- a/udata_front/frontend/menu_helpers.py
+++ b/udata_front/frontend/menu_helpers.py
@@ -10,7 +10,7 @@ def get_current_endpoint(request: Request):
 @front.app_template_global()
 def is_current_endpoint(request: Request, item: Item) -> bool:
     current_endpoint = get_current_endpoint(request)
-    item_endpoint = item.endpoint and item.endpoint.split('.')[:1] 
+    item_endpoint = item.endpoint and item.endpoint.split('.')[:1]
     return request.url_rule.endpoint != "site.home" and (item_endpoint == current_endpoint)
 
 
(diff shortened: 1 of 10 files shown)


maudetes:
It fixes https://github.com/etalab/data.gouv.fr/issues/479? :)

maudetes:
As a note, we should probably update https://support.data.gouv.fr/ headers next.

udata_front/theme/gouvfr/__init__.py
```diff
@@ -47,7 +47,23 @@
...
+            'gouvfr.show_page',
+            args={'slug': 'about/ressources'}
+        ),
+        nav.Item(
+            _('How to publish data ?'),
+            'gouvfr.show_page',
+            args={'slug': 'onboarding/producteurs'}
+        ),
+        nav.Item(
+            _('How to use data ?'),
```
maudetes, line 62: Maybe this instead:
```suggestion
            _('How to use open data ?'),
```

udata_front/frontend/menu_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+
+@front.app_template_global()
+def is_current_page(request: Request, item: Item) -> bool:
+    return is_current_endpoint(request, item) and item.args == request.view_args
+
+
+@front.app_template_global()
+def is_parent_of_current_endpoint(request: Request, items: List[Item]) -> bool:
+    current_item_in_items = map(lambda item: is_current_page(request, item), items)
+    return any(current_item_in_items)
```
maudetes, line 28: I think this would be more pythonic:
```suggestion
    return any(is_current_page(request, item) for item in items)
```

maudetes (commented):
Some minor comments :)

udata_front/theme/gouvfr/__init__.py
nicolaskempf57: French translations are :
- `Commencer sur data.gouv.fr` for the button
- `Qu'est ce que data.gouv.fr ?`
- `Comment publier des données ?`
- `Comment exploiter des données ?`

maudetes: Yes ok, let's keep `how to use data`!
But for the first, I think we can go for `getting started on data.gouv.fr`. :)

nicolaskempf57: Done !

maudetes (approved):
Yay, I think we're good!

PR #118, opened 2022-06-13
Fix criterion 7.5

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/809


maudetes (approved):
Looks good to me! :muscle:

theme/less/components/toast.less
```diff
@@ -129,5 +129,11 @@
     padding: .5em 2em;
     font-family: Avenir, Helvetica, Arial, sans-serif;
     animation-duration: .15s;
-    display: grid;
-}
\ No newline at end of file
+    display: flex;
```
maudetes, line 132: I think we have some issue with hover. See:
image

nicolaskempf57: Done ! This now extends the `.fr-btn` and use DSFR colors

maudetes (approved):
:sparkler:

PR #121, opened 2022-06-14
Add aria-current to breadcrumbs

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/806


maudetes (approved):
Let's go towards 100% accessibility :rocket:

PR #122, opened 2022-06-16
Add quality score in JS search results for demo

maudetes:
Follows https://github.com/opendatateam/udata/pull/2746


maudetes:
Not needed for demo anymore

PR #124, opened 2022-06-16
Remove map related stuff

abulte:
Related https://github.com/opendatateam/udata/pull/2749


maudetes:
Do we want a changelog entry?

PR #126, opened 2022-06-24
Fix 500 on favicon in admin

abulte:
Fix https://github.com/etalab/data.gouv.fr/issues/873


maudetes (commented):
Thank you for this fix!

I think we now have a 404 on favicon. It seems like the theme doesn't correctly extend admin.html.

abulte:
@maudetes about 404: I think this deserves another issue:
- we need to piggy back on https://github.com/opendatateam/udata/pull/2747 to do things cleanly
- we need to cleanup or remove `udata-front/admin.html` which is useless ATM

maudetes (approved):
Thank you!


Indeed, let's fix the 404 on a second iteration. See this issue: https://github.com/etalab/data.gouv.fr/issues/875.

PR #129, opened 2022-06-29
Manifestomy

quaxsze:


maudetes (approved):
Seems all clear to me.

I'm turning it into a draft to prevent merge -> so that it doesn't block us for an udata-front release in the meantime.

PR #135, opened 2022-07-11
Feat/add quality score

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/718


theme/js/components/dataset/search-result.vue
```diff
@@ -45,13 +45,68 @@
...
+                  />
+                  <QualityItem
+                    :passed="quality.spatial_coverage"
+                    :messagePassed='$t("Spatial coverage filled")'
+                    :messageFailed='$t("Spatial coverage not set")'
+                    class="fr-my-1w"
+                  />
+                  <div class="fr-grid-row fr-grid-row--right not-enlarged">
+                    <a
+                      href="https://guides.etalab.gouv.fr/qualite/documenter-les-donnees/"
```
maudetes, line 89: We should probably link to a dedicated *page*. @agarrone I think you made a draft but couldn't find it. Should we go for a *page*? Would we be ready to publish it when this PR goes live?

theme/js/components/dataset/search-result.vue
```diff
@@ -45,13 +45,68 @@
...
+                  <QualityItem
+                    :passed="quality.license"
+                    :messagePassed='$t("License filled")'
+                    :messageFailed='$t("No license set")'
+                    class="fr-my-1w"
+                  />
+                  <QualityItem
+                    :passed="quality.update_frequency && quality.update_fulfilled_in_time"
+                    :messagePassed='$t("Update frequency followed")'
+                    :messageFailed='quality.update_fulfilled_in_time ? $t("Update frequency not followed") : $t("Update frequency not set")'
```
maudetes, line 66: I think `quality.update_fulfilled_in_time` is False even if defined -> always returning "Update frequency not set".

theme/js/components/dataset/search-result.vue
```diff
@@ -45,13 +45,68 @@
...
+                    class="fr-my-1w"
+                  />
+                  <QualityItem
+                    :passed="quality.temporal_coverage"
+                    :messagePassed='$t("Temporal coverage filled")'
+                    :messageFailed='$t("Temporal coverage not set")'
+                    class="fr-my-1w"
+                  />
+                  <QualityItem
+                    :passed="quality.spatial_coverage"
```
maudetes, line 82: This criterion is called `spatial` only.
```suggestion
                    :passed="quality.spatial"
```

maudetes (commented):
Thank you! I find the tooltip integration smooth.

A very first batch of review on quality criteria. I made the comments on the vue component, but they apply on the jinja templates as well :)

Globally, I'm not crazy of the integration on the dataset card layout. Maybe it stands out too much because of the black massive icon and underlined text? I find it sad that we have a 2 lines wrap with the licence now. Ex:
image

theme/js/components/dataset/search-result.vue
agarrone: Propositions to lighten-up the cards : 
- Allow 2 lignes of description only 
- Remove the underline on the score and the licence
- Let the date on its own line (and the score and licence on the same one)
- Hide the filters under an accordion by default 

Regarding your question @maudetes  I'm planning to publish a page on the quality guide. I'll try to draft it today :)

agarrone: See : https://github.com/etalab/guides.etalab.gouv.fr/pull/179

theme/js/components/dataset/search-result.vue
```diff
@@ -45,13 +45,68 @@
             {{ $filters.excerpt(description, 160) }}
           </p>
           <p class="fr-mb-0 text-mention-grey">
+            <Tooltip class="fr-hidden inline-sm">
+                <template #tooltip>
+                  <h5 class="fr-text--sm fr-my-0">{{$t("Metadata quality:")}}</h5>
+                  <QualityItem
```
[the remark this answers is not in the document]
nicolaskempf57: Added ! I put it below the other criterion about description but I can move it.

maudetes (approved):
Nice! I'm looking forward to testing this live :)

I'm adding some minor non-blocking comments

PR #138, opened 2022-07-12
Fix dsfr register error

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/864


maudetes (approved):
Thank you for this nice fix!

PR #140, opened 2022-07-19
Allow to use tag#tagname to refer to the set of datasets/reuses in ma…

freesoul:
Instead of having to refer manually the id/slug of all datasets/reuses in the markdown files, now you can use tag#transport for instance.
Result is as here https://staging.data.public.lu/fr/pages/theme-mobility/

### DIFF
--- a/udata_front/views/gouvfr.py
+++ b/udata_front/views/gouvfr.py
@@ -110,18 +110,39 @@ def get_object(model, id_or_slug):
             pass
     return obj
 
+def get_objects_from_tag(model, tag) -> list:
+    objects = getattr(model, "objects").filter(tags=tag)
+    return [r for r in objects]
 
 @blueprint.route('/pages/<path:slug>/')
 def show_page(slug):
     content, gh_url, extension = get_page_content(slug)
     page = frontmatter.loads(content)
-    reuses = [get_object(Reuse, r) for r in page.get('reuses') or []]
-    datasets = [get_object(Dataset, d) for d in page.get('datasets') or []]
-    reuses = [r for r in reuses if r is not None]
-    datasets = [d for d in datasets if d is not None]
+
+    data = {}
+    models = {"reuses": Reuse, "datasets": Dataset}
+
+    for data_type, model in models.items():
+        data[data_type] = []
+        for dataset_or_reuse in page.get(data_type) or []:
+            if dataset_or_reuse is None:
+                continue
+            dataset_or_reuse = dataset_or_reuse.strip()
+            if dataset_or_reuse[:4] == "tag#":
+                data[data_type] += get_objects_from_tag(model, dataset_or_reuse[4:])
+            else:
+                res = get_object(model, dataset_or_reuse)
+                if res:
+                    data[data_type].append(res)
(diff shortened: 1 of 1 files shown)


udata_front/views/gouvfr.py
```diff
@@ -111,17 +111,40 @@ def get_object(model, id_or_slug):
...
     page = frontmatter.loads(content)
-    reuses = [get_object(Reuse, r) for r in page.get('reuses') or []]
-    datasets = [get_object(Dataset, d) for d in page.get('datasets') or []]
-    reuses = [r for r in reuses if r is not None]
-    datasets = [d for d in datasets if d is not None]
+
+    data = {}
+    models = {"reuses": Reuse, "datasets": Dataset}
+
+    for data_type, model in models.items():
```
maudetes, line 127: I wonder if we could have a get_objects function that encapsulate this logic? For readability of the `show_page` function.

udata_front/views/gouvfr.py
```diff
@@ -111,17 +111,40 @@ def get_object(model, id_or_slug):
     return obj
 
 
+def get_objects_from_tag(model, tag) -> list:
+    objects = getattr(model, "objects").filter(tags=tag)
+    return [r for r in objects]
```
maudetes, line 116: I think  you could directly return `objects`
```suggestion
    return objects
```

udata_front/views/gouvfr.py
```diff
@@ -111,17 +111,40 @@ def get_object(model, id_or_slug):
     return obj
 
 
+def get_objects_from_tag(model, tag) -> list:
+    objects = getattr(model, "objects").filter(tags=tag)
```
maudetes, line 115: We should probably show only `visible` objects. This would ignore archived or private objects.
```suggestion
    objects = getattr(model, "objects").filter(tags=tag).visible()
```

maudetes (commented):
Thank you for this nice contribution!

I like the idea of using a tag in pages. I wonder if it could be interesting to filter on other properties (ex: organization number). But let's keep it simple for now if it's not needed.

Could you add an entry to the [Changelog](https://github.com/etalab/udata-front/blob/master/CHANGELOG.md) please?

And finally, would it be possible to add a test for this logic in [test_static_pages](https://github.com/etalab/udata-front/blob/master/udata_front/tests/test_static_pages.py)?

freesoul:
Sure, will be doing this today or tomorrow :)

freesoul:
For the test I wanted to do like test_page_inject_objects(), but to test it properly I would need to create different DatasetFactory with tags. But I don't see the property tags in this class in udata. I will look further later :-)

maudetes:
Thank you again for your contribution and the updated code :)

> For the test I wanted to do like test_page_inject_objects(), but to test it properly I would need to create different DatasetFactory with tags. But I don't see the property tags in this class in udata. I will look further later :-)

I think you can indeed use a similar logic as `test_page_inject_objects()`. You should actually be able to use the `DatasetFactory` with the `tags` property as follow.
```
DatasetFactory(tags=['tag1', 'tag2'])
```

PR #141, opened 2022-07-20
Add locale support for md pages

freesoul:
Now it is possible to create markdowns under /pages/XX where XX is the locale in the pages repository and expect them to work following the website chosen locale

Example:

Pages:
https://staging.data.public.lu/fr/pages/theme-mobility/ 
https://staging.data.public.lu/en/pages/theme-mobility/

Markdowns:
https://github.com/freesoul/datagouvlu-pages/tree/master/pages/fr 
https://github.com/freesoul/datagouvlu-pages/tree/master/pages/en

### DIFF
--- a/udata_front/views/gouvfr.py
+++ b/udata_front/views/gouvfr.py
@@ -2,7 +2,7 @@ import frontmatter
 import logging
 import requests
 
-from flask import url_for, redirect, abort, current_app
+from flask import url_for, redirect, abort, current_app, g
 from jinja2.exceptions import TemplateNotFound
 from mongoengine.errors import ValidationError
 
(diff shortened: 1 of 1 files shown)


maudetes (commented):
Thank you for this nice PR!

I think @agarrone might be interested for locale support on `pages`.

I'm wondering if we could make some optimization to prevent multiple calls and having different page structure patterns (with or without locale folder):
- We could add locale as an explicit arg in `show_page` and add it in href (ex: `/pages/theme-mobility?lang=fr`). It would prevent multiple calls, in particular if we want to add a fallback on `DEFAULT_LANGUAGE`.
- We could also support only the locale pattern as in https://github.com/freesoul/datagouvlu-pages/tree/master/pages with the `fr`, `en`, etc. folder for all pages. We would need to migrate https://github.com/etalab/datagouvfr-pages/. We would still add a fallback on `DEFAULT_LANGUAGE`.

Once again, I would be interested for your opinions @quaxsze and @abulte.

PR #143, opened 2022-07-24
Switch to poetry

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/892


pyproject.toml
```diff
@@ -0,0 +1,64 @@
...
+    "Topic :: Software Development :: Libraries :: Python Modules",
+    "License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)"
+]
+
+[tool.poetry.dependencies]
+python = ">=3.7,<4"
+feedparser = "6.0.8"
+python-frontmatter = "1.0.0"
+sentry-sdk = {version = "1.1.0", extras = ["flask"]}
+udata = { git = "https://github.com/opendatateam/udata.git", rev = "poetrySwitch" }
```
maudetes, line 29: Don't forget to remove the branch when merging https://github.com/opendatateam/udata/pull/2757

maudetes, line 29: When running `poetry install` in udata-front, it seemed to do a git clone and reinstalled udata. Somehow it doesn't happen anymore, it may have been related to my `.egg-info`.
I'm not sure how it works to use the local udata instead of the one defined in this dependency?

Do you think we should specify [a local package](https://python-poetry.org/docs/dependency-specification/#path-dependencies)?
Don't know how the dependency would work on a packaged udata-front though.

For reference, we had this specific logic: https://github.com/etalab/udata-front/blob/c60b9d3ca0f5236f7f9c3a92c07d7e4116226f22/setup.py#L14

maudetes (commented):
See relevant comments I made on https://github.com/opendatateam/udata/pull/2757.

We need to specify instructions on the good practice to have a shared dev env.
I personally manually activated the env (located in ~/.cache/pypoetry/virtualenvs/) in udata-front.
We could also create an env in-project in udata and use it as well in udata-front.
In any case we need to document it I think.

I had some issues when first installing (see my comment), but then it seemed to work smoothly, with both udata and udata-front being in development mode with hot reload working.

quaxsze:
Trying to get an advice on our issue: https://github.com/python-poetry/poetry/issues/6706

PR #145, opened 2022-08-03
fix(dataset): :adhesive_bandage: fix discussion text wrap

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/882


maudetes (approved):
:ok_hand:

PR #152, opened 2022-08-24
Udata ApiV2 changes

quaxsze:


maudetes (commented):
Somehow the parser fails to parse parameters from url correctly

udata_front/views/base.py
```diff
@@ -96,10 +97,10 @@ class SearchView(Templated, BaseView):
     page_size: Optional[int] = None
```
ifteneiftene, line 97: m

PR #154, opened 2022-08-30
Clear specific cache after object update

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/144

### DIFF
--- a/udata_front/theme/gouvfr/templates/dataset/display.html
+++ b/udata_front/theme/gouvfr/templates/dataset/display.html
@@ -24,7 +24,7 @@
 {% set external_source = dataset.extras['remote_url'] %}
 
 {% block extra_head %}
-{% cache cache_duration, 'dataset-extra-head', dataset.id|string, g.lang_code %}
+{% cache cache_duration, 'dataset-extra-head', dataset.id|string, g.lang_code, dataset.last_modified|string %}
 <link rel="canonical" href="{{ url_for('datasets.show', dataset=dataset) }}" />
 <link rel="alternate" type="application/json+oembed" href="{{ url_for('api.oembed', url=dataset.external_url) }}"
     title="{{ dataset.title | urlencode }}" />
@@ -32,14 +32,14 @@
 {% endblock %}
 
 {% block breadcrumb %}
-{% cache cache_duration, 'dataset-breadcrumb', dataset.id|string, g.lang_code %}
+{% cache cache_duration, 'dataset-breadcrumb', dataset.id|string, g.lang_code, dataset.last_modified|string %}
 <li><a href="{{ url_for('datasets.list') }}">{{ _('Datasets') }}</a></li>
 <li class="active" aria-current="page">{{ dataset.acronym or dataset.title|truncate(128) }}</li>
 {% endcache %}
 {% endblock %}
 
 {% block content %}
-{% cache cache_duration, 'dataset-content', dataset.id|string, g.lang_code, current_user.slug or 'anonymous' %}
+{% cache cache_duration, 'dataset-content', dataset.id|string, reuses.page|string, g.lang_code, current_user.slug or 'anonymous', dataset.last_modified|string %}
 
 {{ breadcrumb(self,
     breadcrum_class=breadcrum_class,
--- a/udata_front/theme/gouvfr/templates/dataset/display.html
+++ b/udata_front/theme/gouvfr/templates/dataset/display.html
@@ -39,7 +39,7 @@
 {% endblock %}
 
 {% block content %}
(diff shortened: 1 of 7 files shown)


udata_front/views/dataset.py
```diff
@@ -139,3 +142,16 @@ def group_resources_by_type(resources):
...
+@Dataset.on_update.connect
+def clear_cache_on_updated_dataset(dataset):
+    user = current_user.slug or 'anonymous'
+
+    for lang_code in current_app.config['LANGUAGES'].keys():
+        head_cache_key = make_template_fragment_key("dataset-extra-head", vary_on=[str(dataset.id), lang_code])
+        cache.delete(head_cache_key)
+        breadcrumb_cache_key = make_template_fragment_key("dataset-breadcrumb", vary_on=[str(dataset.id), lang_code])
+        cache.delete(breadcrumb_cache_key)
+        content_cache_key = make_template_fragment_key("dataset-content", vary_on=[str(dataset.id), "1", lang_code, user])
```
quaxsze, line 156: The template cache here is defined in the jinja file as following:
`{% cache cache_duration, 'dataset-content', dataset.id|string, reuses.page|string, g.lang_code, current_user.slug or 'anonymous' %}`

The `reuses.page` is not know in the event context. Should we remove this key? Knowing the keys `dataset-content`, `dataset.id`, `g.lang_code` and `current_user.slug` might be enough?

maudetes (commented):
Don't forget the changelog :p

maudetes (approved):
Nice!, well done for the deep dive into flask caching!

It seemed to work nicely locally :)

udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -39,7 +39,7 @@
 {% endblock %}
 
 {% block content %}
-{% cache cache_duration, 'dataset-content', dataset.id|string, reuses.page|string, g.lang_code, current_user.slug or 'anonymous' %}
+{% cache cache_duration, 'dataset-content', dataset.id|string, g.lang_code, current_user.slug or 'anonymous' %}
```
abulte, line 42: I guess `reuses.page|string` was invalidating the cache when a reuse or added or deleted to the dataset, isn't interesting to keep?

abulte:
Dumb question: isn't adding the modified date of an object to the cache string enough to do the job?

udata_front/theme/gouvfr/templates/dataset/display.html
maudetes: Oooh I thought the same, but it's actually the reuse page number. **It is indeed necessary for pagination**.

The same key is probably needed on reuse and orga pages to paginate the different elements.

maudetes:
> Dumb question: isn't adding the modified date of an object to the cache string enough to do the job?

Ahah good point. I don't see why it wouldn't work.

udata_front/theme/gouvfr/templates/dataset/display.html
quaxsze: The reuses.page is not know in the event context. We cannot recreate such a cache key in the signal.
I do not understand your comment @maudetes  :/

maudetes: > The reuses.page is not know in the event context. We cannot recreate such a cache key in the signal.

Maybe using the modified date as @abulte suggested would do it. We wouldn't need to manually invalidate the cache here.
If it somehow doesn't work, we would need to iterate on all reuses pages (as you do with lang) to invalidate the cache for all reuses pages.

> I do not understand your comment @maudetes :/

I'm talking about the paginating inside a dataset, reuse or orga page, such as https://www.data.gouv.fr/fr/datasets/base-sirene-des-entreprises-et-de-leurs-etablissements-siren-siret/?reuses_page=2#community-reuses. If you don't use the current page number in the key, you would be unable to paginate. The previous link would return the result for the `reuses_page=1`

udata_front/theme/gouvfr/templates/organization/display.html
```diff
@@ -17,11 +17,14 @@
...
 
 {% block breadcrumb %}
+{% cache cache_duration, 'org-breadcrumb', org.id|string, g.lang_code, org.last_modified|string %}
 <li><a href="{{ url_for('organizations.list') }}">{{ _('Organizations') }}</a></li>
 <li class="active" aria-current="page">{{ org.name }}</li>
+{% endcache %}
 {% endblock %}
 
 {% block main_content %}
+{% cache cache_duration, 'org-content', org.id|string, g.lang_code, current_user.slug or 'anonymous', org.last_modified|string %}
```
maudetes, line 27: I did not test, but I think you'll need to add a key such as `reuses.page|string` and `datasets.page|string`, to allow for pagination, ex:https://www.data.gouv.fr/fr/organizations/etalab/?datasets_page=2#organization-datasets

PR #95, opened 2022-03-22
Feat/update search pages

nicolaskempf57:
Left to decide :

- [x] Do we keep the temporal cover filter ? Or do we remove it and add a publication date filter ? We remove it.
- [x] What do we do on mobile ? Filters are in a sidemenu with an accordion-like style on mobile.
- [x] Quickfix : Enlever le survol sur les résultats des jdds pour gagner de la place
- [x] Unifier l'affichage du nb de résultats sur les pages
- [x] Orga : pb de conteneur
- [x] décaler le nb de résultats au dessus des résultats et remonter les filtres

Close https://github.com/etalab/data.gouv.fr/issues/774
Close https://github.com/etalab/data.gouv.fr/issues/775
Close https://github.com/etalab/data.gouv.fr/issues/347
Close https://github.com/etalab/data.gouv.fr/issues/610

PR #156, opened 2022-09-06
Fix/i18n errors

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/903

We could refactor our locale handling in a new PR to lazy load the messages only for the current lang : https://vue-i18n.intlify.dev/guide/advanced/lazy.html

This also remove `vue3-datepicker`, an unnecessary dependency since https://github.com/etalab/udata-front/pull/95.


package.json
```diff
@@ -60,8 +60,7 @@
     "process": "^0.11.10",
     "rimraf": "^3.0.2",
     "stylemark": "^3.1.7",
-    "vue-i18n-extract": "^2.0.4",
-    "vue3-datepicker": "^0.3.2"
```
abulte, line 64: Is this related to this PR?

nicolaskempf57: Not really, it was unused since #95

I can rollback and do this in a new PR.

abulte: No need I think, a message on the PR should be enough!

abulte (approved):

package.json
nicolaskempf57: Done !

PR #158, opened 2022-09-12
Fix lighthouse reported errors

nicolaskempf57:
This implements some fixes to https://github.com/etalab/data.gouv.fr/issues/894


udata_front/theme/gouvfr/templates/footer.html
```diff
@@ -75,52 +73,54 @@ <h3 class="fr-footer__top-cat">{{ _('Social networks') }}</h3>
...
+                            <li>
                                 <a
                                     href="https://pypi.org/project/udata/{{ package_version('udata') }}/"
-                                    title="{{ _('udata deployed version: {version}').format(
-                                        version=package_version('udata')
-                                    ) }}"
                                     class="fr-footer__top-link"
                                 >
-                                    {{ package_version('udata') }}
+                                    {{ _('udata deployed version: {version}').format(
```
maudetes, line 104: Maybe we could use a simpler wording like `udata version`?

udata_front/frontend/helpers.py
```diff
@@ -342,18 +342,15 @@ def i18n_alternate_links():
...
         if request.args:
             params.update(request.args)
         if request.view_args:
             params.update(request.view_args)
 
         for lang in current_app.config['LANGUAGES']:
-            if lang != current_lang:
-                url = url_for(request.endpoint, lang_code=lang, **params)
-                links.append(LINK_PATTERN.format(url=url, lang=lang))
+            url = url_for(request.endpoint, lang_code=lang, **params, _external=True)
```
maudetes, line 352: The `test_i18n_alternate_links` test should be updated to reflect the `_external=True` option + the current page alternate.

udata_front/theme/gouvfr/templates/footer.html
```diff
@@ -75,52 +73,54 @@ <h3 class="fr-footer__top-cat">{{ _('Social networks') }}</h3>
...
                                 </a>
                             </li>
                             <li>
                                 <a
                                     href="https://github.com/etalab/udata-front/"
-                                    title="{{ _('data.gouv.fr extension: udata-front') }}"
                                     class="fr-footer__top-link"
                                 >
-                                    udata-front
+                                    {{ _('data.gouv.fr extension: udata-front') }}
```
maudetes, line 114: How about something like `udata theme/front extension: udata-front`?

maudetes (commented):
Nice! :)

I think we can add some stronger assertions on Lighthouse CI job following these fixes.
We can make sure we're at 100% on accessibility and best practices. See https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#assert

nicolaskempf57:
> I'm not 100% convinced about the new `Technical information` part in the footer, but I'm making some wording suggestions.

This is a way to keep the same information and make it compliant but I agree that this makes it way bigger than before. 

We can see with the team if it's something we want to keep this way or find a new way to show it like a dedicated page.

maudetes:
For info, a snapshot of these Technical Information
image

maudetes:
I think we can remove direct link to version release, simplifying to :
image

We could keep the previous column layout with DINUM + etalab logo with 2 lines (one for udata and the other udata-front).
What do you think?

nicolaskempf57:
> I think we can remove direct link to version release, simplifying to :

Done and it's not in Lighthouse report anymore.

I agree it's better with the previous layout. I didn't add the title "technical informations" from the new one but I can.

maudetes (approved):

agarrone:
Sorry I forgot to check this out but your solution in fine. 
Thanks!

PR #159, opened 2022-09-12
Feat/add captchetat

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/822


quaxsze:
:rotating_light: Merge this first in order to work https://github.com/opendatateam/udata/pull/2766

udata_front/frontend/__init__.py
```diff
@@ -77,3 +77,10 @@ def test_500():
     if app.config.get('DEBUG_TOOLBAR'):
         from flask_debugtoolbar import DebugToolbarExtension
         DebugToolbarExtension(app)
+
+    # Security override init
+    from udata.auth import init_security
+    from udata_front.forms import ExtendedAgainRegisterForm
```
abulte, line 83: `ExtendedRegisterForm` (as a name) should be enough now that we have this one in udata-front?

udata_front/frontend/__init__.py
```diff
@@ -79,8 +79,9 @@ def test_500():
         DebugToolbarExtension(app)
 
     # Security override init
-    from udata.auth import init_security
+    from udata.auth import security
     from udata_front.forms import ExtendedRegisterForm
 
-    init_security(
-        app, confirm_register_form=ExtendedRegisterForm, register_form=ExtendedRegisterForm)
+    with app.app_context():
```
maudetes, line 85: Nice! :)

udata_front/settings.py
```diff
@@ -33,3 +33,14 @@
...
+# CAPTCHEtat
+CAPTCHETAT_BASE_URL = 'https://sandbox-api.piste.gouv.fr/piste/captcha'
+CAPTCHETAT_OAUTH_BASE_URL = 'https://sandbox-oauth.piste.gouv.fr'
+CAPTCHETAT_OAUTH_TOKEN_URL = f'{CAPTCHETAT_OAUTH_BASE_URL}/api/oauth/token'
+CAPTCHETAT_GET_CAPTCHA_URL = f'{CAPTCHETAT_BASE_URL}/simple-captcha-endpoint'
+CAPTCHETAT_VALIDATION_URL = f'{CAPTCHETAT_BASE_URL}/valider-captcha'
+CAPTCHETAT_CLIENT_ID = '565d420d-1321-467b-93f4-f86f702c08ca'
+CAPTCHETAT_CLIENT_SECRET = '04260b77-195c-471c-b4f6-539098cc4817'
+TOKEN_CACHE_KEY = 'captchetat-bearer-token'
+TOKEN_DEFAULT_MEDIATYPE = 'application/octet-stream'
```
maudetes, line 46:
```suggestion
CAPTCHETAT_TOKEN_CACHE_KEY = 'captchetat-bearer-token'
CAPTCHETAT_TOKEN_DEFAULT_MEDIATYPE = 'application/octet-stream'
```

maudetes, line 46: Do we still need `TOKEN_DEFAULT_MEDIATYPE`?

udata_front/frontend/__init__.py
```diff
@@ -77,3 +77,11 @@ def test_500():
...
         from flask_debugtoolbar import DebugToolbarExtension
         DebugToolbarExtension(app)
+
+    # Security override init
+    from udata.auth import security
+    from udata_front.forms import ExtendedRegisterForm
+
+    with app.app_context():
+        security._state.register_form = ExtendedRegisterForm
+        security._state.confirm_register_form = ExtendedRegisterForm
```
maudetes, line 87: Do we need both `confirm_register_form` and `register_form`?

maudetes (commented):
Nice! It seems to work well! :clap:

I have some additional comments on the code :)

udata_front/api.py
```diff
@@ -0,0 +1,68 @@
...
+    def get(self):
+        '''CaptchEtat endpoint for captcha generation and validation'''
+        args = captchetat_parser.parse_args()
+        headers = {'Authorization': 'Bearer ' + bearer_token()}
+        captchetat_url = current_app.config.get('CAPTCHETAT_GET_CAPTCHA_URL')
+        req = requests.get(captchetat_url, headers=headers, params=args)
+        accept = request.accept_mimetypes.copy()
+        if args['get'] == "sound":
+            accept.append(("audio/*", 1))
+        if args['get'] == "p":
```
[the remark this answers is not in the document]
nicolaskempf57: Based on CaptchEtat doc, it could be `html, layout-stylesheet, script-include, image, reload-icon, sound-icon, reload-disabled-icon, sound-disabled-icon, sound, p`

udata_front/settings.py
nicolaskempf57: I don't think

udata_front/settings.py
```diff
@@ -33,3 +33,14 @@
 GUIDES_ORGANIZATION_URL = 'https://guides.etalab.gouv.fr/data.gouv.fr/creer-rejoindre-organisation/'
 GUIDES_DATASET_URL = 'https://guides.etalab.gouv.fr/data.gouv.fr/publier-jeu-de-donnees/'
 GUIDES_REUSE_URL = 'https://guides.etalab.gouv.fr/reutilisation/publier-reutilisation/'
+
+# CAPTCHEtat
+CAPTCHETAT_BASE_URL = 'https://sandbox-api.piste.gouv.fr/piste/captcha'
```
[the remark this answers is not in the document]
nicolaskempf57: We can change the security form only if `CAPTCHETAT_BASE_URL` or all variables are set so it's not blocking for udata-front setups outside of ours.

udata_front/frontend/__init__.py
nicolaskempf57: Both are required because one or the other is used depending of `SECURITY_CONFIRMABLE` value. This was already fixed for udata https://github.com/opendatateam/udata/pull/2588

udata_front/api.py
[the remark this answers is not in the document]
maudetes: all these choices! :p ok thank you!

udata_front/settings.py
```diff
@@ -33,3 +33,11 @@
...
```
maudetes, line 38: I think we should document the behaviour of `CAPTCHETAT_BASE_URL` to activate or deactivate captchetat.
We should also add a comment on additional `CAPTCHETAT_CLIENT_ID` and `CAPTCHETAT_CLIENT_SECRET`.

Maybe adding it in settings.py + in changelog would do :)

udata_front/api.py
```diff
@@ -0,0 +1,74 @@
...
+    log.debug(f'New access token requested from {url}')
+    try:
+        oauth = requests.post(url, data={
+            'grant_type': 'client_credentials',
+            'scope': 'WRITE',
+            'client_id': current_app.config.get('CAPTCHETAT_CLIENT_ID'),
+            'client_secret': current_app.config.get('CAPTCHETAT_CLIENT_SECRET')
+        })
+        body = oauth.json()
+        access_token = body.get('access_token', "")
```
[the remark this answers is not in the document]
nicolaskempf57: I added a condition to prevent it

udata_front/api.py
```diff
@@ -0,0 +1,74 @@
...
+        return previous_value
+    log.debug(f'New access token requested from {url}')
+    try:
+        oauth = requests.post(url, data={
+            'grant_type': 'client_credentials',
+            'scope': 'WRITE',
+            'client_id': current_app.config.get('CAPTCHETAT_CLIENT_ID'),
+            'client_secret': current_app.config.get('CAPTCHETAT_CLIENT_SECRET')
+        })
+        body = oauth.json()
```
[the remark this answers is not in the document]
nicolaskempf57: raise_for_status() added ! There is now a try block around both `bearer_token` call and captchetat call.

udata_front/settings.py
nicolaskempf57: Done !

udata_front/settings.py
```diff
@@ -33,3 +33,16 @@
...
+
+# CAPTCHEtat
+
+# You can activate CaptchEtat, a captcha.com integration by providing
+# CAPTCHETAT_BASE_URL, CAPTCHETAT_OAUTH_BASE_URL, CAPTCHETAT_CLIENT_ID and CAPTCHETAT_CLIENT_SECRET
+CAPTCHETAT_BASE_URL = None
+CAPTCHETAT_OAUTH_BASE_URL = None
+CAPTCHETAT_CLIENT_ID = None
+CAPTCHETAT_CLIENT_SECRET = None
+CAPTCHETAT_OAUTH_TOKEN_URL = f'{CAPTCHETAT_OAUTH_BASE_URL}/api/oauth/token'
```
maudetes, line 45: Sadly `CAPTCHETAT_OAUTH_TOKEN_URL` uses `CAPTCHETAT_OAUTH_BASE_URL` defined above, not the one loaded in the current_app (and possibly overridden). We could concatenate both in `api.py` instead of formatting it here? Or defining the pattern there?

udata_front/api.py
```diff
@@ -0,0 +1,80 @@
...
+@ns.route('/', endpoint='captchetat')
+class CaptchEtatAPI(API):
+    @apiv2.expect(captchetat_parser)
+    @apiv2.doc('captchetat')
+    def get(self):
+        '''CaptchEtat endpoint for captcha generation and validation'''
+        args = captchetat_parser.parse_args()
+        try:
+            token = bearer_token()
+            if token:
```
maudetes, line 61: We should probably abort if not token as well?

nicolaskempf57: We could reverse the condition here : https://github.com/etalab/udata-front/pull/159/files#diff-5b34f309ae9d500b0f3685c5942e88c1832756ce5021a57282768cd2301fa08cR44 and throw if there is no access_token

maudetes: Indeed! It's not blocking since now it will fail if token is missing in header, but it would be better to raise the error as soon as possible indeed :)

maudetes (approved):
:ok_hand:

PR #161, opened 2022-10-04
Switch to pyproject.toml

quaxsze:


maudetes (commented):
Almost there :)

I sadly have an issue with udata deps. We can discuss a strategy to bypass this problem.

I think you should clean pre-commit and tox.ini, similar to what you did in udata :)

maudetes:
Closing for now. See https://github.com/etalab/data.gouv.fr/issues/939#issuecomment-1281163133

PR #168, opened 2022-11-04
Feat/use harvest metadata

maudetes:
Follows https://github.com/opendatateam/udata/pull/2762.

Display harvest metadata in an html toggle next to extras toggle.
image

Show dates on top of dataset page and dataset card even for harvested datasets.
Use `last_update`, now based on more recent date for resources (modified, published and harvest modified date): https://github.com/opendatateam/udata/pull/2762#discussion_r1014046875.

This is a minimal step to mitigate the impact of separating the harvest dates, before dealing with https://github.com/etalab/data.gouv.fr/issues/856

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Display and use new harvest metadata [#168](https://github.com/etalab/udata-front/pull/168)
 
 ## 2.0.13 (2022-11-02)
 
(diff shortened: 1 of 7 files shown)


nicolaskempf57 (commented):
`external_source` is still used here, in both the condition and href, as a variable instead of a function : https://github.com/etalab/udata-front/pull/168/files#diff-d5e0edc6387330398745e23afaaa276b2986de431e4589cc4db452a45903f541R223

I can't make the suggestion with GH interface because this part didn't change.

nicolaskempf57 (approved):

PR #171, opened 2022-11-29
fix: :adhesive_bandage: fix banner links with 2 underlines

nicolaskempf57:
Require for https://github.com/etalab/data.gouv.fr/issues/959


maudetes (approved):
Good catch!

PR #172, opened 2022-11-29
Fix text for datasets button in org page

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/792

### DIFF
--- a/udata_front/theme/gouvfr/templates/organization/display.html
+++ b/udata_front/theme/gouvfr/templates/organization/display.html
@@ -105,7 +105,7 @@
             href="{{ url_for('datasets.list', organization=org.id) }}"
             class="fr-btn fr-btn--secondary fr-icon-eye-line fr-btn--icon-left"
         >
-            {{ ngettext('See {0} dataset', 'See {0} datasets', total_datasets) }}
+            {{ ngettext('See %(num)d dataset', 'See %(num)d datasets', total_datasets) }}
         </a>
     </div>
 </section>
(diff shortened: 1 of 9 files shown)


udata_front/theme/gouvfr/templates/organization/display.html
```diff
@@ -105,7 +105,7 @@ <h2>{{ _('Datasets') }}<sup>{{ total_datasets }}</sup></h2>
             href="{{ url_for('datasets.list', organization=org.id) }}"
             class="fr-btn fr-btn--secondary fr-icon-eye-line fr-btn--icon-left"
         >
-            {{ _('See %(count)s datasets', count=total_datasets ) }}
+            {{ ngettext('See {0} dataset', 'See {0} datasets', total_datasets) }}
```
maudetes, line 108: From what I understand from ngettext, we should use `%(num)d`: https://pythonhosted.org/Flask-BabelEx/#flask.ext.babelex.Domain.ngettext

maudetes (commented):
Thank you for the modification!

Don't forget to re-extract translation ;)
I'm wondering about the singular string, should it be "See the dataset" instead of "See {0} dataset" (thinking of the French translation as well!

nicolaskempf57:
> Thank you for the modification!
> 
> Don't forget to re-extract translation ;) I'm wondering about the singular string, should it be "See the dataset" instead of "See {0} dataset" (thinking of the French translation as well!

Done !

maudetes (approved):
:ok_hand:

PR #173, opened 2022-11-29
Fix Api Representation for media

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/985

### DIFF
--- a/udata_front/api.py
+++ b/udata_front/api.py
@@ -78,14 +78,13 @@ class CaptchEtatAPI(API):
             return json.loads(req.content)
         request.accept_mimetypes = MIMEAccept(accept)
 
-        values_gen = request.accept_mimetypes.values()
-        result_image = next((True for value in values_gen if 'image/' in value), False)
-        values_gen = request.accept_mimetypes.values()
-        result_sound = next((True for value in values_gen if 'audio/' in value), False)
+        values_gen = list(request.accept_mimetypes.values())
+        result_image = any('image/' in value for value in values_gen)
+        result_sound = any('audio/' in value for value in values_gen)
 
         if result_image or result_sound:
             resp = make_response(bytes(req.content))
-            resp.headers["Content-Type"] = "image/*" if result_image else "audio/*"
+            resp.headers['Content-Type'] = 'image/*' if result_image else 'audio/*'
             return resp
 
         return req.content
--- a/udata_front/api.py
+++ b/udata_front/api.py
@@ -73,7 +73,7 @@ class CaptchEtatAPI(API):
 
         if args['get'] in ['image', 'sound']:
             resp = make_response(bytes(req.content))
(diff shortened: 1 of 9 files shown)


udata_front/api.py
```diff
@@ -77,11 +77,15 @@ def get(self):
...
-        return req.content
 
-    @apiv2.representation('image/*')
-    @apiv2.representation('audio/*')
-    def media(data, code, headers):
-        resp = make_response(bytes(data), code)
-        resp.headers.extend(headers)
-        return resp
+        values_gen = request.accept_mimetypes.values()
+        result_image = next((True for value in values_gen if 'image/' in value), False)
```
maudetes, line 82: Could we do something like:
`any('image/' in value for value in values_gen)`

udata_front/api.py
```diff
@@ -77,11 +77,15 @@ def get(self):
...
+        values_gen = request.accept_mimetypes.values()
+        result_sound = next((True for value in values_gen if 'audio/' in value), False)
+
+        if result_image or result_sound:
+            resp = make_response(bytes(req.content))
+            resp.headers["Content-Type"] = "image/*" if result_image else "audio/*"
+            return resp
```
maudetes, line 89: If we return a content with `make_response`, it doesn't get process anymore by flask right?

maudetes (commented):
Don't forget the changelog?

udata_front/api.py
quaxsze: Yes :)

udata_front/api.py
```diff
@@ -77,11 +77,14 @@ def get(self):
...
         request.accept_mimetypes = MIMEAccept(accept)
-        return req.content
 
-    @apiv2.representation('image/*')
-    @apiv2.representation('audio/*')
-    def media(data, code, headers):
-        resp = make_response(bytes(data), code)
-        resp.headers.extend(headers)
-        return resp
+        values_gen = list(request.accept_mimetypes.values())
```
maudetes, line 81: Not sure we need to meddle with `request.accept_mimetypes` anymore.
Could we directly use `make_response` based on `args['get']` being image or audio?

maudetes (commented):
Seems to work perfectly, but I'm making a small suggestion

quaxsze:
@nicolaskempf57 We remove the `api.representation` line for `media`. The whole APIv2 was passing through it.
Instead we use the `get` argument to know which header to include in the response. Does it look okey to you?

nicolaskempf57:
> @nicolaskempf57 We remove the `api.representation` line for `media`. The whole APIv2 was passing through it. Instead we use the `get` argument to know which header to include in the response. Does it look okey to you?

Yes ! This was done because the captcha calls doesn't include any `Accept` header so flask-rest-x treats it as `application/json` and that prevent it for sending the response as bytes.

maudetes (approved):
Great! :clap: 

Don't forget to update the tests

nicolaskempf57 (approved):

PR #174, opened 2022-12-05
Fix rgaa criterion 8.9

nicolaskempf57:
This PR goal is to prevent `<div>` for being used as `<p>`.
This is a first pass on the site. I'm sure I have missed some cases and I'm sorry because the review will be painful !

Close https://github.com/etalab/data.gouv.fr/issues/811


maudetes:
Due to the changes on the dataset page, could you merge before I make a review! :pray: Sorry for the delay

udata_front/theme/gouvfr/templates/home.html
```diff
@@ -90,8 +90,8 @@ <h3>{% include theme('svg/logotype.svg') %} {{ _('is') }}</h3>
             <div class="fr-grid-row justify-between fr-mt-md-9v fr-mt-2w">
                 {% for (label, metric) in metrics %}
                 <div class="filet fr-my-1v fr-my-md-0">
-                    <div class="fs-hg f-bold text-grey-400">{{ current_site.metrics[metric]|format_number }}</div>
-                    <span class="text-grey-400">{{ label }}</span>
+                    <p class="fr-m-0 fr-h1 text-grey-400">{{ current_site.metrics[metric]|format_number }}</p>
+                    <p class="fr-m-0 text-grey-400">{{ label }}</p>
```
maudetes, line 94: Shouldn't it be in the same `<p>`?

maudetes (approved):
Thank you for this pass!
Seems like most issues encountered at audit time have already been fixed by using DSFR components, which is nice!

I've tried taking a look at new components we've introduced since, but did not see any new bad patterns, but I haven't gotten through the entire site though

PR #178, opened 2022-12-14
Fix rgaa criterion 9.2

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/812


udata_front/theme/gouvfr/templates/post/display.html
```diff
@@ -101,7 +101,11 @@ <h2 class="fr-h2">{{ _('Actions') }}</h2>
 
 {% if previous_post or next_post %}
 <section class="fr-container fr-my-7w fr-py-3w border-top">
-    <div class="fr-grid-row fr-grid-row--gutters justify-between">
+    <nav
+        role="navigation"
+        aria-label="{{_("Navigate in the page collection")}}"
```
maudetes, line 106: Don't forget to extract the translation ;)

Maybe it should be:
```suggestion
        aria-label="{{_("Navigate in the post collection")}}"
```

maudetes (approved):
Thank you! It seems to match the recommendations.
I hope we're not missing issues that did not appear before though

PR #196, opened 2023-01-19
Fix all js scripts waiting for DOMContentLoaded (already passed)

nicolaskempf57:
Following https://github.com/etalab/udata-front/pull/183, the js is loaded after the `DOMContentLoaded` event. The scripts wait for this `DOMContentLoaded` event, that appended before them so they are never triggered.

Fix https://github.com/etalab/data.gouv.fr/issues/1031
Fix https://github.com/etalab/udata-front/issues/197
Also fix copy buttons on current dataset page, sort on reuses and organizations pages.

### DIFF
--- a/theme/js/components/vanilla/accordion.js
+++ b/theme/js/components/vanilla/accordion.js
@@ -29,29 +29,25 @@ The accordion panel needs to have :
 
 import { easing, tween, styler } from "popmotion";
 
-export default (() => {
-  document.addEventListener("DOMContentLoaded", () => {
-    const togglers = document.querySelectorAll("[data-accordion-button]");
+const togglers = document.querySelectorAll("[data-accordion-button]");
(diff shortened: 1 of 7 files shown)


theme/js/components/vanilla/accordion.js
```diff
@@ -29,29 +29,25 @@ The accordion panel needs to have :
 
 import { easing, tween, styler } from "popmotion";
 
-export default (() => {
-  document.addEventListener("DOMContentLoaded", () => {
-    const togglers = document.querySelectorAll("[data-accordion-button]");
+const togglers = document.querySelectorAll("[data-accordion-button]");
+console.log(togglers)
```
maudetes, line 33:
```suggestion
```

maudetes (approved):
Thank you for the fix! :pray:

PR #201, opened 2023-01-26
Compute dates depending on harvest dict

quaxsze:


udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -243,7 +243,11 @@ <h2 class="subtitle">{{ _('Temporality') }}</h2>
                                 {% endif %}
                                 <dt class="fr-text--sm fr-my-0 text-mention-grey">{{ _('Creation date') }}</dt>
                                 <dd class="fr-text--sm fr-ml-0 fr-mt-0 fr-mb-2w">
-                                    {{ dataset.created_at|dateformat(format='long') }}
+                                    {% if dataset.harvest and dataset.harvest.created_at %}
+                                        {{ [dataset.created_at, dataset.harvest.created_at]|max|dateformat(format='long') }}
```
maudetes, line 247: I don't think it should be the max for created_at. More like take the harvest date if set, else the internal one (as defined in https://github.com/etalab/data.gouv.fr/issues/1006)

udata_front/theme/gouvfr/templates/dataset/resource/card.html
```diff
@@ -3,7 +3,7 @@
 {% set available = resource.extras['check:status']
     and resource.extras['check:status'] >= 200 and resource.extras['check:status'] < 400 %}
 {% set unavailable = resource.extras['check:status'] and resource.extras['check:status'] >= 400 %}
-{% set last_update = resource.published if resource.published > resource.modified else resource.modified %}
+{% set last_update = resource.harvest.modified_at if (resource.harvest and resource.harvest.modified_at > resource.modified) else resource.modified %}
```
maudetes, line 6: `resource.harvest` may be defined and `resource.harvest.modified_at` be None. Thus leading to `TypeError: '>' not supported between instances of 'NoneType' and 'datetime.datetime'`

quaxsze:
Tests will fail until https://github.com/opendatateam/udata/pull/2815 is merge in master

maudetes (approved):
I like the renaming strategy better, it implies much fewer modifications on udata-front side, and only simplification. I think we're all clear on this side, well done!

PR #204, opened 2023-02-03
opens new tab for external portal banner link in ressources

SiamoTuttiAntiMarco:
opens the external portal in a new tab at click on the link, when the dataset is flag as coming from an external source.

example : https://www.data.gouv.fr/fr/datasets/projet-de-loi-de-finances-pour-2016-plf-2016-jaune-personnels-affectes-dans-les-cabinets-ministeriels/

It depends on your preferences... well it's mine


ThibaudDauce:
Right now we'll keep the current way, you can always ctrl-clic to open a new tab (but you cannot undo a `target="_blank"`)

PR #205, opened 2023-02-08
Make newsletter url configurable

maudetes:
Following https://github.com/etalab/data.gouv.fr/issues/1017

I think we can change to the future url in our instance configuration :)
We could also set it as an empty string in  `settings.py` instead.

### DIFF
--- a/udata_front/settings.py
+++ b/udata_front/settings.py
@@ -44,3 +44,6 @@ CAPTCHETAT_OAUTH_BASE_URL = None
 CAPTCHETAT_CLIENT_ID = None
 CAPTCHETAT_CLIENT_SECRET = None
 CAPTCHETAT_TOKEN_CACHE_KEY = 'captchetat-bearer-token'
+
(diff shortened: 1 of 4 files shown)


udata_front/settings.py
```diff
@@ -44,3 +44,6 @@
 CAPTCHETAT_CLIENT_ID = None
 CAPTCHETAT_CLIENT_SECRET = None
 CAPTCHETAT_TOKEN_CACHE_KEY = 'captchetat-bearer-token'
+
+# Newsletter
+NEWSLETTER_SUBSCRIPTION_URL = 'https://infolettres.etalab.gouv.fr/subscribe/rn7y93le1'
```
maudetes, line 49:
```suggestion
NEWSLETTER_SUBSCRIPTION_URL = 'https://f.info.data.gouv.fr/f/lp/infolettre-data-gouv-fr-landing-page/lk3q01y6'
```

PR #206, opened 2023-02-08
Show dsfr notice banner without JS

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/1013

This is based on the new notice component from DSFR 1.6.
It's backported here to prevent a DSFR update in this PR.

It uses the data-fr-js attribute that is added on DSFR JavaScript boot.

### DIFF
--- a/udata_front/theme/gouvfr/templates/header.html
+++ b/udata_front/theme/gouvfr/templates/header.html
@@ -206,7 +206,7 @@
             <div class="fr-notice__body">
                 <p class="fr-notice__title">
                     {{
-                        _("This is a degraded experience of {site}. Please enable JavaScript and use a modern browser.")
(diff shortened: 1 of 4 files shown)


udata_front/theme/gouvfr/templates/header.html
```diff
@@ -199,3 +199,18 @@
...
     </div>
 </header>
+
+<div data-show-no-js>
+    <div class="fr-notice fr-notice--warning">
+        <div class="fr-container">
+            <div class="fr-notice__body">
+                <p class="fr-notice__title">
+                    {{
+                        _("This is a degraded experience of {site}. Please enable JavaScript and use a modern browser.")
```
maudetes, line 209: Do we want to add "up to date" ? :p

maudetes (approved):
Nice! :star2: kinda feels like magic

PR #215, opened 2023-03-02
Use vue-toaster fork

nicolaskempf57:
Close https://github.com/etalab/udata-front/security/dependabot/7
Close https://github.com/etalab/udata-front/security/dependabot/12
Close https://github.com/etalab/udata-front/security/dependabot/9
Close https://github.com/etalab/udata-front/security/dependabot/8

There are also 4 other vulnerabilities that was related to it but they require updates of other packages.


package.json
```diff
@@ -35,6 +33,7 @@
   },
   "devDependencies": {
     "@conciergerie.dev/select-a11y": "^3.1.0",
+    "@conciergerie.dev/vue-toaster": "^2.3.6",
```
maudetes, line 36: It should be in `dependencies` instead of `devDependencies` right?

maudetes (approved):
Nice! You've applied the patch that was here directly to the fork?

nicolaskempf57:
> Nice! You've applied the patch that was here directly to the fork?

Yes, I forgot the `component` change for accessibility. I'm adding it to v2.4.0.

PR #220, opened 2023-03-08
Front modifications according to schema new url field

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1052
fix https://github.com/etalab/data.gouv.fr/issues/1018


udata_front/views/dataset.py
```diff
@@ -161,6 +161,9 @@ def group_resources_by_schema(resources):
     """Group a list of `resources` by `schema`"""
     groups = dict()
     for resource in resources:
-        if len(resource.schema) > 0 and resource.schema['name'] not in groups:
-            groups[resource.schema['name']] = resource.schema
+        try:
```
maudetes, line 164: Shouldn't we test `if 'name' in resource.schema` instead of a try catch?

theme/js/components/dataset/resource/resource.vue
```diff
@@ -146,7 +146,7 @@
                   <div class="fr-grid-row fr-grid-row--middle fr-mb-1w">
                     <p class="fr-text--xs fr-m-0">{{$t('This file is following a schema: ')}}</p>
                     <p class="fr-tag fr-tag--sm fr-icon-checkbox-circle-line fr-tag--icon-left">
-                      {{resource.schema.name}}
+                      {{resource.schema.name ? resource.schema.name : resource.schema.url}}
```
maudetes, line 149: There is also a `See schema documentation` button later on that should be displayed only when resource.schema.name :)

maudetes (commented):
:pray:

maudetes (approved):

PR #224, opened 2023-03-13
Fix of the login form after Flask-Security and WTForms upgrade

quaxsze:
Related to https://github.com/etalab/data.gouv.fr/issues/1053


udata_front/theme/gouvfr/templates/macros/forms.html
```diff
@@ -7,13 +7,14 @@
 {% elif f.type == 'SubmitField' %}
 <button type="submit" class="fr-btn">{{ submit }}</button>
 {% elif f.type == 'BooleanField' %}
-<div class="fr-checkbox-group">
-    {{ f() }}
-    <label class="fr-label" for="{{ f.id }}">
-        {{ f.label.text }}
-    </label>
-</div>
+{{ render_boolean_field(f) }}
```
maudetes, line 10: Did you take this update from Flask Security forms?

maudetes (commented):
Don't forget the changelog ;)

I think the security override init may not work anymore. `captcha_code` attribute is missing when trying to display http://dev.local:7000/fr/register

udata_front/theme/gouvfr/templates/macros/forms.html
quaxsze: no I made a func for it:
```
{% macro render_boolean_field(f) %}
<div class="fr-checkbox-group">
    {{ f() }}
    <label class="fr-label" for="{{ f.id }}">
        {{ f.label.text }}
    </label>
</div>
{% endmacro %}
```
I did this so i can reorganize the login form after the update

maudetes (approved):
Seems like we're clear! :clap: 
Don't forget the changelog!

PR #229, opened 2023-03-21
Fix dataset page on responsive

nicolaskempf57:
This fix the page having blank on the right due to tooltip.

This also removes some old media-queries pre-dsfr and refactor some code to use new queries.


maudetes (approved):
:ok_hand: Don't forget the changelog entry ;)

PR #230, opened 2023-03-21
New organization page

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/1061

Note : Depending on the search used, the page could show empty datasets because they are not filtered out of mongo search. The tests are updated due to this change.


udata_front/theme/gouvfr/templates/organization/display.html
```diff
@@ -31,180 +43,322 @@
...
+    </div>
+    <div
+        id="information-panel"
+        class="fr-tabs__panel"
+        role="tabpanel"
+        aria-labelledby="information"
+        tabindex="0"
+    >
+        <div class="fr-container">
+        {% call() information_section(_('Description')) %}
```
maudetes, line 310: We shouldn't have the description once again here? It already appears in `presentation` panel

udata_front/theme/gouvfr/templates/organization/display.html
```diff
@@ -31,180 +43,322 @@
...
+        {% endcall %}
+        {% call(slot) information_title_with_button(_('Members')) %}
+            {% if slot == "button" %}
                 {% set member = org.member(current_user) %}
                 {% if not member %}
                     {% set pending_request = org.pending_request(current_user) %}
                     {% if not pending_request and not read_only_mode %}
-                    <li class="fr-my-2w vuejs">
+                    <div class="fr-my-2w vuejs">
                         <request-membership orga='{{org.id}}'>
```
maudetes, line 320: this prop has been renamed I think
```suggestion
                        <request-membership organization='{{org.id}}'>
```

maudetes (commented):
WIP review! I like the layout a lot :clap:
Seems nice that we could reuse the search components almost as is!

udata_front/theme/gouvfr/templates/organization/display.html
nicolaskempf57: You're right ! I don't know what Figma I was looking at but it's not here anymore

maudetes (approved):

PR #237, opened 2023-04-05
Integration of MCP login

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1067


udata_front/settings.py
```diff
@@ -46,5 +46,14 @@
...
 CAPTCHETAT_TOKEN_CACHE_KEY = 'captchetat-bearer-token'
 
+# MonComptePro
+
+# You can activate MonComptePro SSO, a moncomptepro.beta.gouv.fr login integration
+MONCOMPETPRO_OPENID_CONF_URL = None
+MONCOMPETPRO_SCOPE = None
+MONCOMPETPRO_CLIENT_ID = None
+MONCOMPETPRO_CLIENT_SECRET = None
+MONCOMPETPRO_REDIRECT_URL = None
```
maudetes, line 56: We don't use it, do we?

udata_front/views/mcp.py
```diff
@@ -0,0 +1,42 @@
...
+from udata.i18n import I18nBlueprint
+from udata_front.frontend import oauth
+from udata.models import datastore
+from udata.auth import login_user
+
+
+blueprint = I18nBlueprint('mcp', __name__, url_prefix='/mcp')
+
+
+@blueprint.route('/login', endpoint='login')
```
maudetes, line 11: What about logout? Should we do anything when logging out of udata or MCP ?

udata_front/views/mcp.py
```diff
@@ -0,0 +1,42 @@
...
+def auth():
+    client = oauth.create_client('mcp')
+    if not client:
+        abort(404)
+
+    token = client.authorize_access_token()
+    mcp_user = token.get('userinfo')
+    if not mcp_user:
+        mcp_user = client.userinfo()
+    user = datastore.find_user(email=mcp_user['email'])
```
maudetes, line 31: It makes it really simple that we use email as id between our model and MCP :clap:

maudetes (commented):
It seems to work nicely! :muscle: I have some questions but nothing blocking.
Would be nice to have @rdubigny take a look at this if possible :)

udata_front/theme/gouvfr/templates/security/login_user.html
```diff
@@ -41,7 +41,19 @@ <h1 class="fr-h4 fr-mb-0">{{ _('Log in') }}</h1>
                             </p>
                             {% endif %}
                         </div>
-                </form>
+                    </form>
+                    <form class="form" method="get" action="{{ url_for('mcp.login') }}">
```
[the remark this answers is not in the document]
maudetes: > The button should also be displayed on the register form.

maudetes (commented):
Feedback from @rdubigny that seemed content :)

> I would have added a a test to check that a user that doesn't have a paswword can't login without one

maudetes (approved):
:clap: :clap:

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Current (in progress)
 
+- MonComptePro SSO integration [#237](https://github.com/etalab/udata-front/pull/237):
+    - New button on login and register page
```
maudetes, line 6: Maybe just mention that it's optional, if some configs are set?

maudetes (approved):
:clap: 
Seems nice ! :)

PR #239, opened 2023-04-17
Add version in js chunk file name

maudetes:
Applies to _commonjsHelpers only in our case.

It allows us to version this file, preventing using a previous cached version that has changed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
           command: |
             virtualenv venv
             source venv/bin/activate
-            pip install setuptools==66.1.1
+            pip install --force-reinstall setuptools==66.1.1
             pip install pip-tools
       - run:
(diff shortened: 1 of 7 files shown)


nicolaskempf57 (approved):

maudetes:
`_commonjsHelpers-legacy.js` being loaded in raw.html, we need to find a way to load it with its correct name (manifest, package version, etc.)

maudetes (commented):
We've opted for using the package version in the chunk file names, thus having the `[name].[version].js` pattern.
We use bumpr to keep the version number in sync when releasing.
We've added an option in theme_static to use this pattern instead of the default dummy param version (`?_=version`).

Additionally, I've had an issue with setuptools version being installed. I'm unclear why, but it seems that it doesn't use the specify version if we don't do a force-reinstall.

PR #241, opened 2023-04-24
Handle previous discussion urls

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/1078


maudetes (approved):
Thank you for this fix! :pray:

It would be nice to have some tests on correct hash behavior, but we can do that in a second iteration

PR #244, opened 2023-04-25
Move theme to gouvfr

nicolaskempf57:
This PR moves the `theme` folder from udata-front root to `gouvfr` theme folder.

It also adds some glob to ease the use of udata-front with another theme.

Close https://github.com/etalab/data.gouv.fr/issues/565


nicolaskempf57:
> I wonder if we should rename the second `theme` directory in `udata_front/theme/${theme}/theme/...` for clarity?

I agree, I think we can rename it to `assets`. What do you think ?

crowdin.yml
```diff
@@ -2,5 +2,5 @@ files:
   - source: /udata_front/theme/gouvfr/translations/gouvfr.pot
     translation: >-
       /udata_front/theme/gouvfr/translations/%two_letters_code%/LC_MESSAGES/%file_name%.po
-  - source: /theme/js/locales/en.json
```
maudetes, line 5: We'll need to update vue-i18n-extract.config.js as well

maudetes (approved):
Aside from one comment, seems to work nicely :)

crowdin.yml
nicolaskempf57: Because `vue-i18n-extract` wants a glob and we must also match `udata-front-plugins-helper`, I had to slightly change the match so it's targeting any `.js` or `.vue` file in a `js` folder in any folder in `udata_front` and not in `udata_front/theme`. I can change the `js`folder check to `assets` too but I don't know if it's better.

maudetes: Seems fine!

PR #259, opened 2023-05-23
Fix tooltip accessibility

nicolaskempf57:
This improves the accessibility of our tooltip implementation, adds a new toggletip component and uses the toggletip in place of existing tooltips.

Close https://github.com/etalab/data.gouv.fr/issues/924


maudetes (approved):
Seems to work nicely :)

For later, do you think that we still need to use the hack we [introduced over there](https://github.com/etalab/udata-front/blob/master/CHANGELOG.md#2011-2022-09-02) on z-index results ?

PR #265, opened 2023-06-12
Fix tab sequence

nicolaskempf57:
This PR prevents tabbing inside a file component when it's hidden.

It also locks the read-more button at the bottom of its container.

Close https://github.com/etalab/data.gouv.fr/issues/1087


udata_front/theme/gouvfr/assets/js/components/utils/read-more.vue
```diff
@@ -11,8 +11,10 @@ A very simple component that display its content up to a maximum height and then
 -->
 
 <template>
-  <div ref="container" class="read-more" :class="{expand: expanded}" :style="{height: containerHeight + 'px'}">
-    <slot></slot>
+  <div class="relative">
+    <div ref="container" class="read-more" :class="{expand: expanded}" :style="{height: containerHeight + 'px'}">
```
maudetes, line 15: I think you may need to change the height of the shading, it cuts a bit too soon when reading more.
Ex [on the SIREN](http://dev.local:7000/fr/datasets/base-sirene-des-entreprises-et-de-leurs-etablissements-siren-siret/)
image

maudetes (commented):
Thank you, the tabulation seems to work well on resources now, it's actually quite satisfying :)

udata_front/theme/gouvfr/assets/js/components/utils/read-more.vue
nicolaskempf57: I forgot to update the CSS after the HTML changes. Done.

maudetes (approved):
:pray:

PR #269, opened 2023-06-19
Fix header search

nicolaskempf57:
DSFR adds a `overflow: hidden` to `fr-container-lg--fluid` so it hides our popup from showing.


maudetes (approved):
:pray:

PR #273, opened 2023-06-26
Add captcha on forgot password page

quaxsze:


maudetes (commented):
When having `SECURITY_RETURN_GENERIC_RESPONSES=True`, it returns `Instructions to reset your password have been sent to ..`, even if Captcha Fails (and mail hasn't been sent).

We should first check for the captcha result before returning the generic sentence.

quaxsze:
PR closed due to the degradation of the user's xp when using captcha with flask-security generic responses

PR #289, opened 2023-08-16
Fix padding and background

nicolaskempf57:
Closes https://github.com/etalab/data.gouv.fr/issues/1124

### DIFF
--- a/udata_front/theme/gouvfr/assets/less/components/table.less
+++ b/udata_front/theme/gouvfr/assets/less/components/table.less
@@ -2,7 +2,7 @@
 
 .fr-table--no-background {
     thead {
-        background: none;
+        background-color: transparent;
     }
     th {
         min-width: 7rem;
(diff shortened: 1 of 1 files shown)


udata_front/theme/gouvfr/assets/less/components/table.less
```diff
@@ -22,3 +22,14 @@
...
     --table-offset: 0;
 }
+
+.fr-table:has(thead th .fr-btn) {
+    tbody td {
+        .fr-py-2w;
+        .fr-px-5v;
+    }
+    th {
+        padding: 0.75rem 0.5rem;
```
maudetes, line 32: I think this get overridden?
I'm still having alignment issue as shown in issue I think.

maudetes (commented):
It's better with the line separation in between the header and body indeed :pray:

udata_front/theme/gouvfr/assets/less/components/table.less
nicolaskempf57: Not working in Firefox, I'll find another way

nicolaskempf57: This is done with a new class `fr-table--dense` now

maudetes (approved):
:ok_hand:

PR #290, opened 2023-08-16
Add guides to the menu

nicolaskempf57:
Closes https://github.com/etalab/data.gouv.fr/issues/1145


maudetes (approved):
As an open question, I wonder if we should use a `target="_blank"`, because the behaviors differs from the others items?
It could show an icon and open the guides in a new tab?
Not sure if it would be better in terms of accessibility though

nicolaskempf57:
> It could show an icon and open the guides in a new tab?

Here is an example of a menu link with a `target="_blank"` :

Xnapper-2023-08-21-10 18 51

If we add a title indicating that the link will open in a new tab, it should be fine. At least, this is the DSFR way of doing it.

maudetes:
What do you think @agarrone?

agarrone:
Not a big fan on this particular case

PR #291, opened 2023-08-16
Use the word file instead of resource

nicolaskempf57:
See https://github.com/etalab/data.gouv.fr/issues/929


maudetes (approved):
bye bye resource :hole:

PR #298, opened 2023-09-11
fix: :arrow_up: update deps

nicolaskempf57:
Closes https://github.com/etalab/udata-front/pull/277

### DIFF
--- a/udata-front-plugins-helper/package.json
+++ b/udata-front-plugins-helper/package.json
@@ -19,15 +19,18 @@
     "publish-stable": "npm publish --access public",
     "publish-next": "npm run publish-stable -- --tag next"
   },
+  "overrides": {
+    "tough-cookie": "^4.1.3"
+  },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


udata-front-plugins-helper/package.json
```diff
@@ -19,15 +19,18 @@
     "publish-stable": "npm publish --access public",
     "publish-next": "npm run publish-stable -- --tag next"
   },
+  "overrides": {
+    "tough-cookie": "^4.1.3"
```
maudetes, line 23: Not sure I totally understand how overrides work, but can't we remove the `overrides` once package-lock has been updated?

maudetes (approved):
:pray:

udata-front-plugins-helper/package.json
nicolaskempf57: I put it here to match the fix we did on [udata-front](https://github.com/etalab/udata-front/blob/master/package.json#L23) and so it doesn't rely on package-lock not being erased.

PR #300, opened 2023-09-12
Fix dependencies according to udata's Flask-Babel migration

quaxsze:


maudetes (changes requested):
Thank you! However, I think we still need to make translations work with jinja templates

maudetes (approved):
Jinja templates trads have been fixed in https://github.com/opendatateam/udata/pull/2897

PR #301, opened 2023-09-13
Don't cache links to other languages

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/1166


maudetes (approved):
Thank you for the quickfix© :pray:

PR #304, opened 2023-09-28
Add discoverability indicator form

nicolaskempf57:
Close https://github.com/etalab/data.gouv.fr/issues/1133

### DIFF
--- a/udata_front/theme/gouvfr/assets/js/components/vanilla/dialog.js
+++ b/udata_front/theme/gouvfr/assets/js/components/vanilla/dialog.js
@@ -1,0 +1,63 @@
+/**
+ *
+ * @param {HTMLElement} dialog
+ */
+function showDialog (dialog) {
+  if (dialog) {
+    dialog.classList.remove("fr-hidden");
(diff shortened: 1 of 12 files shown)


nicolaskempf57:
Do we also remove the header button ?

udata_front/theme/gouvfr/assets/js/components/vanilla/dialog.js
```diff
@@ -0,0 +1,63 @@
...
+  window.localStorage.setItem(dialog.id, "true");
+};
+
+/**
+ *
+ * @param {HTMLElement} dialog
+ * @param {number} triggerCount
+ * @param {string} startsWithString
+ */
+function triggerDialog(dialog, triggerCount = 1, startsWithString = '/') {
```
maudetes, line 28: I would increment triggerCount to 2, so that it displays on third page being visited?

maudetes (approved):
Seems to work nicely! :clap: 

Forwarding the question to @agarrone for the current header button

PR #309, opened 2023-10-27
Update deps

nicolaskempf57:
This PR removes the existing package-lock.json to recreate it from the package.json.

This is done because we had some dependencies like `stylus` that are still in our package-lock, not required anymore by udata-front but marked as optional in some of our dependencies. 

This creates a state that prevent NPM from removing it with CLI commands.

Removing the package-lock triggers an update of all our dependencies following semver. It works great for all packages except the DSFR. We're now checking DSFR version in the format [`1.9.x`](https://github.com/npm/node-semver#tilde-ranges-123-12-1) instead of [`1.x.y`](https://github.com/npm/node-semver#caret-ranges-123-025-004). It seems that the `1.10` introduce some breaking changes for our use case.

Closes https://github.com/etalab/udata-front/pull/305
Closes https://github.com/etalab/udata-front/pull/306
Closes https://github.com/etalab/udata-front/pull/308

Edit : I revert the Node version change. It was necessary for the DSFR 1.10 but we're not using it for now.


README.md
```diff
@@ -125,7 +125,7 @@ inv serve
 
 ##### 🏗 Installing the javascript dependencies
 
-First, you need to use [Node](https://nodejs.org/) (version 14+) on your platform. You should consider [installing NVM](https://github.com/creationix/nvm#installation) which uses the existing .nvmrc.
+First, you need to use [Node](https://nodejs.org/) (version 20+) on your platform. You should consider [installing NVM](https://github.com/creationix/nvm#installation) which uses the existing .nvmrc.
```
maudetes, line 128: 16+?

maudetes (approved):
Not sure I understand why the node version 20 fails. But nice upgrades :pray: 
(don't forget the changelog)

PR #313, opened 2023-11-17
Moving reuse metrics hook before discussion section

maudetes:
New layout on DVF reuse page:
image


maudetes:
Should we use `fr-btn--sm` to make the button the same size as the discussion ones? And what about the download button on other pages (dataset, orga, dashboard)?

agarrone:
Yes the same size sounds better :)

maudetes:
Made in https://github.com/opendatateam/udata-metrics/pull/8/commits/2ab424c997dc99d0f0b0a29faeddc80ba8a791ab

PR #321, opened 2023-11-22
Ignore resize observer loop

nicolaskempf57:


maudetes (approved):
I'm not sure I understand why this error is triggered. Could you try to explain it shortly please ? :pray:
What about other errors, such as `ResizeObserver is not defined` ?

nicolaskempf57:
ResizeObserver is used by the `useElementSize` function in our [read-more](https://github.com/etalab/udata-front/blob/master/udata_front/theme/gouvfr/assets/js/components/ReadMore/ReadMore.vue#L42) component.

It allows us to watch the content height and decide to trigger the read-more or not. It also tells us the correct height to use if the read-more content is smaller than our default read-more height.

The error is quite obscure but the class works as expected as per [the spec repo](https://github.com/WICG/resize-observer/issues/38).

The root cause is that we are receiving more than one notification before each frame. An option we can try is to throttle the changes on our side like [here](https://github.com/mui/material-ui/blob/94b4c4b57e1e0fef757d59da405c3b5fb508220e/packages/mui-base/src/TextareaAutosize/TextareaAutosize.tsx#L197) and see if it fixes it.

I reproduce this error on a Safari 16.0 and it's caused on launch, maybe because of [the internal initial value](https://github.com/WICG/resize-observer/issues/8).

Regarding the errors `ResizeObserver is not defined` and `Can't find variable: ResizeObserver`, they are triggered on older browsers ([not updated since 2019](https://caniuse.com/resizeobserver)). We can provide a polyfill but the main ones are unmaintained so we'll have to support it ourself.

PR #322, opened 2023-11-27
Fixed ol not appearing in descriptions

Jorek57:
Close https://github.com/etalab/data.gouv.fr/issues/1083


maudetes (approved):
:rocket: Thank you!

nicolaskempf57 (approved):
Nice ! In the future, if / when we remove `sanitize.css`, we should update these to use DSFR defaults.

PR #336, opened 2023-12-21
Call udata schema endpoint

nicolaskempf57:
Requires https://github.com/opendatateam/udata/pull/2882

### DIFF
--- a/udata_front/theme/gouvfr/assets/js/components/SchemaSelect/SchemaSelect.vue
+++ b/udata_front/theme/gouvfr/assets/js/components/SchemaSelect/SchemaSelect.vue
@@ -41,8 +41,8 @@ export default defineComponent({
   setup(props, {emit}) {
     /** @type {Promise<Array<import("../../types").MultiSelectOption>>} */
     const initialOptions = getCatalog().then(catalog => catalog.map(schema => ({
-      label: schema.name,
-      value: schema.name,
+      label: schema.id,
+      value: schema.id,
     })));
     const change = (value) => emit("change", value);
     return {
--- a/udata_front/theme/gouvfr/assets/js/api/schemas.js
+++ b/udata_front/theme/gouvfr/assets/js/api/schemas.js
@@ -1,5 +1,5 @@
-import axios from "axios";
 import config from "../config";
+import { api } from "../plugins/api";
 
 /**
  * A schema version.
@@ -13,9 +13,9 @@ import config from "../config";
  * A json schema associated with a resource.
  * @typedef Schema
  * @type {object}
- * @property {string} name - The schema name.
- * @property {string} schema_type - The schema type.
- * @property {Array<SchemaVersion>} versions - The schema versions.
+ * @property {string} id - The schema name.
+ * @property {string} label - The schema label for humans.
+ * @property {Array<string>} versions - The schema versions.
  */
 
 /**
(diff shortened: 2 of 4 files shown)


udata_front/theme/gouvfr/assets/js/components/SchemaSelect/SchemaSelect.vue
```diff
@@ -41,8 +41,8 @@ export default defineComponent({
   setup(props, {emit}) {
     /** @type {Promise<Array<import("../../types").MultiSelectOption>>} */
     const initialOptions = getCatalog().then(catalog => catalog.map(schema => ({
-      label: schema.name,
-      value: schema.name,
+      label: schema.id,
+      value: schema.id,
```
ThibaudDauce, line 45: I think this change broke the frontend… We should keep `.name` here. 
image

ThibaudDauce (dismissed):

udata_front/theme/gouvfr/assets/js/components/SchemaSelect/SchemaSelect.vue
nicolaskempf57: The PR is not working right now and is waiting for https://github.com/etalab/udata-front/pull/378 to be merged (for the [api file](https://github.com/etalab/udata-front/pull/378/files#diff-e9f6af4ac65a05931acf1a1beb71e93ce4ddbfa3e920bf2d7af0514e738ec127) in datagouv-components).

Then, a new version of the package is necessary to allow schema api changes to be reflected in udata-front.

udata_front/theme/gouvfr/datagouv-components/src/api/schemas.ts
```diff
@@ -49,14 +47,10 @@ let catalogRequest: Promise<Array<RegisteredSchema>> | null = null;
...
  */
 export function getCatalog() {
-  if(!config.schema_catalog_url) {
-    catalogRequest = Promise.resolve([]);
-  }
   if (catalogRequest) {
     return catalogRequest;
   }
-  catalogRequest = axios.get<SchemaResponseData>(config.schema_catalog_url)
+  catalogRequest = api.get<SchemaResponseData>("datasets/schemas/")
```
maudetes, line 53: I think we can remove `schema_catalog_url` / `SCHEMA_CATALOG_URL` config, right?

maudetes (commented):
Nice that we can merge this now :)
This will finally close https://github.com/datagouv/data.gouv.fr/issues/1148 completely.

udata_front/theme/gouvfr/datagouv-components/src/api/schemas.ts
nicolaskempf57: I removed the `schema_catalog_url` const and the check to `SCHEMA_CATALOG_URL` in metadata file. 

Do we want to remove it from udata-front settings too ? If we do it, it should be added to all our environments `cfg` because it's still used in udata with a default setting as `None`.

maudetes: I think we should remove it since udata-front doesn't use it! Our dev/demo/prod env have the value set properly in their cfg.
I don't know if front-kit env would be impacted though :)

nicolaskempf57: front-kit get the variable from the `.env` file at build time 👍🏻 
I'm adding a note to the PR description about the setting change for devs and I'll merge it.

maudetes (approved):
I think we're good, thank you! :pray:

PR #337, opened 2023-12-22
Add lazy loading to md image

nicolaskempf57:
Closes https://github.com/etalab/data.gouv.fr/issues/1188

This PR also introduces a unified `<markdown-viewer>`.

### DIFF
--- a/udata_front/theme/gouvfr/templates/dataset/display.html
+++ b/udata_front/theme/gouvfr/templates/dataset/display.html
@@ -103,8 +103,9 @@
             </h1>
             <h2 id="description" class="subtitle fr-mb-1w">{{ _('Description') }}</h2>
             <read-more max-height="description" class="fr-mb-n1w">
-                {{ dataset.description|markdown }}
+                <markdown-viewer data-content="{{dataset.description|to_json}}"></markdown-viewer>
             </read-more>
+            <noscript>{{ dataset.description | markdown }}</noscript>
         </section>
         <section class="fr-col-12 fr-col-md-4" data-read-more-max-height="description">
             {% if sysadmin %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- force image lazy loading to markdown content [#337](https://github.com/etalab/udata-front/pull/337)
 
 ## 3.2.12 (2023-12-15)
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "udata-front",
-  "version": "3.2.11.dev",
+  "version": "3.2.13.dev",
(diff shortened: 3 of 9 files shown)


udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -103,8 +103,9 @@ <h1 class="fr-mb-5v fr-h3">
             </h1>
             <h2 id="description" class="subtitle fr-mb-1w">{{ _('Description') }}</h2>
             <read-more max-height="description" class="fr-mb-n1w">
-                {{ dataset.description|markdown }}
+                <markdown-viewer data-content="{{dataset.description|to_json}}"></markdown-viewer>
```
maudetes, line 106: If we keep this idea, we should apply the same logic for other markdown contents? In particular, I'm thinking of *pages*.

maudetes (commented):
I could not make it work locally, markdown-viewer does not seem to get mounted.

The issue I see with this approach is that before vue app can be loaded (can take up some time), description will be empty. I think it may degrade user experience?

udata_front/theme/gouvfr/templates/dataset/display.html
nicolaskempf57: This is indeed the idea to have a unified viewer for all contents. This could allow us to use DSFR tables everywhere for example instead of our custom ones.

nicolaskempf57:
> I could not make it work locally, markdown-viewer does not seem to get mounted.

I'll look into it.

> The issue I see with this approach is that before vue app can be loaded (can take up some time), description will be empty. I think it may degrade user experience?

Vue loading shouldn't be longer here than on other pages but I agree that nothing was done currently regarding performance and we should start working on it or at least planning some work like Sentry Performance Monitoring.

maudetes:
> > I could not make it work locally, markdown-viewer does not seem to get mounted.
> 
> I'll look into it.

My bad, I had not build the correct udata-front assets. After rebuilding, it works correctly!

udata_front/theme/gouvfr/templates/dataset/display.html
ThibaudDauce, line 106: I think it's better to do a simple escape the HTML attribute (if it's possible, I didn't test it but I don't see why it wouldn't work… See https://stackoverflow.com/questions/9187946/escaping-inside-html-tag-attribute-value
```suggestion
                <markdown-viewer content="{{dataset.description|replace('"', '&quot;')}}"></markdown-viewer>
```
This allow to remove all the `ref` and `onMounted` inside the component and only keep a single props `content`.

nicolaskempf57: We tried to do it with a `props` for the Search component and this is indeed simpler for us. But the unintended side-effect is that an invalid props will break vue and the component will never mount. I'll try it when we redo it to use the same parser as the markdown editor.

PR #343, opened 2024-01-24
Fix vue runtime

nicolaskempf57:
Reverts `package-lock.json` changes.

#324 should prevent bumpr changes in upcoming versions.

Also it seems that there is somehow a bug with vue runtime in version `3.4.15` and how we load our components.


quaxsze (approved):

maudetes (approved):
Thank you!

PR #361, opened 2024-02-20
Fix schema manipulation

ThibaudDauce:
See https://github.com/opendatateam/udata/pull/2949

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -264,6 +264,9 @@ jobs:
       - attach_workspace:
           at: .
       - run:
+          name: Prepare udata.cfg file with udata-front plugin
+          command: cp .lighthouseci/front-plugin-udata.cfg udata.cfg
+      - run:
           name: Generate udata fixtures
           command: |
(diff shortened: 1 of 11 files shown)


maudetes (commented):
I think there is still an issue for the Resource in Vue: https://github.com/etalab/udata-front/blob/5b63e7bf20944fedc3f42fa52b94a405f5f45f59/udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/ResourceAccordion.vue#L390

ThibaudDauce:
> I think there is still an issue for the Resource in Vue:
> 
> https://github.com/etalab/udata-front/blob/5b63e7bf20944fedc3f42fa52b94a405f5f45f59/udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/ResourceAccordion.vue#L390

I think in Vue there is no problem to access a dict or a class like this (since dict and classes are the same thing in JSON)

nicolaskempf57 (approved):

PR #362, opened 2024-02-20
Hide quality score for specific harvest backends

maudetes:
This is a proposal to start harvesting some plateforms (ex with https://github.com/opendatateam/udata/pull/2957) without penalizing them due to harvest issues.
Indeed, if quality errors shows prominently, some producers won't want to be harvested.
We could remove this when we're sure the quality issues are no more the consequences of the harvest mapping.

It would show like this on a Dataset page:
image
Detailed items are still available in the toggle, but not warning shown directly.

On the Dataset card, the quality metadata would simply not be shown.
image

What do you think @agarrone?

### DIFF
--- a/udata_front/settings.py
+++ b/udata_front/settings.py
@@ -77,3 +77,6 @@ NEWSLETTER_SUBSCRIPTION_URL = 'https://f.info.data.gouv.fr/f/lp/infolettre-data-
 
 # Metrics API
 METRICS_API = None
+
+# Metadata quality is hidden for datasets harvested from theses backends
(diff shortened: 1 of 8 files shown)


agarrone:
Thanks.
How long do you think it would take to get those metadata right ?

maudetes:
> How long do you think it would take to get those metadata right ?

It depends whether the means of action are on our side or on the side of the harvested platforms I think. Probably a few months.

agarrone:
Allright I think we need to say somehow that we are working on improving this. Maybe just add "We are working on the issue".

PR #363, opened 2024-02-22
Fix CI

nicolaskempf57:


maudetes (approved):
I think this is still valid, right?

PR #371, opened 2024-03-06
Add admin datasets page

nicolaskempf57:
Closes https://github.com/etalab/data.gouv.fr/issues/1287

This PR adds the dataset table page to the new administration. It doesn't include the statistics part, because there is no easy way to include them in vue-land for now. I think it's big enough to be in a separate PR.

This PR requires https://github.com/datagouv/udata-front/pull/466 but testing https://github.com/datagouv/udata-front/pull/466 is difficult outside of this PR.

To test this PR with a working environment, you should build the updated version of datagouv-components (added to this branch too) and link it instead of using the version from the package.json :
```
# build package
cd udata_front/theme/gouvfr/datagouv-components
npm ci
npm run build
cd ../../../../ # return to udata-front

# link package
npm link ./udata_front/theme/gouvfr/datagouv-components
```

~~Note : The dsfr update seems to break the full-page tabs and some colors are not ours anymore.~~

Note 2 : Seems to work now


package.json
```diff
@@ -87,6 +86,7 @@
   },
   "license": "LGPL-2.1-only",
   "dependencies": {
+    "@gouvfr/dsfr": "^1.11.1",
```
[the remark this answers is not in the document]
nicolaskempf57: Yes, you can use `npm ci --force` to test it for now but @etalab/data.gouv.fr-components will require a dsfr update.

 I would prefer to have all component PRs merged and a clean release pipeline in place for the package before doing it to be able to better test the components with the new dsfr version.

.storybook/preview-head.html
```diff
@@ -6,6 +6,7 @@
 <meta name="api-root" content="https://dev.data.gouv.fr/api/1/" />
```
maudetes, line 6: Don't forget the changelog ;)

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,191 @@
...
+            <span class="fr-icon--sm fr-icon-star-s-line" aria-hidden="true"></span>
+            <template #tooltip> {{ t('Favorites') }}</template>
+          </Tooltip>
+        </AdminTableTh>
+      </tr>
+    </thead>
+    <tbody>
+      <tr v-for="dataset in datasets" :key="dataset.id">
+        <AdminTableTd>
+          <AdminContentWithTooltip>
```
maudetes, line 82: I'm not sure I've understood how AdminContentWithTooltip is used?

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,191 @@
...
+        <AdminTableTh scope="col">{{ t('Score') }}</AdminTableTh>
+        <AdminTableTh scope="col">
+          <Tooltip>
+            <span class="fr-icon--sm fr-icon-chat-3-line" aria-hidden="true"></span>
+            <template #tooltip> {{ t('Discussions') }}</template>
+          </Tooltip>
+        </AdminTableTh>
+        <AdminTableTh
+          :sortable="true"
+          @sort="(direction) => updateSort('views', direction)"
```
maudetes, line 40: Should we sort by `-views` on first click instead of ascending on the metrics columns?
The sorting order would then differ between columns though.
In current admin, seems like we're always sorting in descending order (title, dates, metrics).
cc @agarrone

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,191 @@
...
+      <tr v-for="dataset in datasets" :key="dataset.id">
+        <AdminTableTd>
+          <AdminContentWithTooltip>
+            <a class="fr-link" :href="getDatasetLinkToAdmin(dataset)">
+              {{ dataset.title }}
+            </a>
+          </AdminContentWithTooltip>
+        </AdminTableTd>
+        <AdminTableTd>
+        <AdminBadge :type="dataset.private ? 'default' : 'info'">{{ dataset.private ? t('Private') : t('Public') }}</AdminBadge>
```
maudetes, line 89: I think status could also be archived or deleted. See the logic used [in the current admin](https://github.com/opendatateam/udata/blob/f1a3b68e803b87bc0ba9282d2468de2204b89ad8/js/components/datatable/cells/visibility.vue#L8).

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,191 @@
...
+import { admin_root } from '../../../config';
+import { summarize } from '../../../helpers';
+import { DatasetSortedBy, SortDirection } from '../../../types';
+
+const emit = defineEmits<{
+  (event: 'sort', sort: string): void
+}>();
+
+defineProps<{
+  datasets: Array<Dataset | DatasetV2>
```
maudetes, line 144: Shouldn't we use either one or the other?

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,191 @@
...
+const metrics = ref<Record<string, number>>({});
+const sort = ref<SortDirection>('desc');
+const sortedBy = ref<DatasetSortedBy>('created');
+
+const { t } = useI18n();
+
+function updateSort(column: DatasetSortedBy, direction: SortDirection) {
+  sort.value = direction;
+  sortedBy.value = column;
+  emit('sort', `${sort.value === 'asc' ? "" : "-"}${sortedBy.value}`);
```
maudetes, line 156: Do we have to use signals? I find it hard to follow

udata_front/theme/gouvfr/assets/js/config.ts
```diff
@@ -85,6 +85,8 @@ export const unchecked_types = _jsonMeta("unchecked-types");
 export const api_root = _meta("api-root");
 export const api_2_root = _meta("api-2-root");
 
+export const metrics_api_root = _meta("metrics-api");
```
maudetes, line 88: I wish we would not need to query metrics api in admin. We should probably add it as dataset metrics instead.

udata_front/theme/gouvfr/assets/js/pages/Admin/Admin.vue
```diff
@@ -19,72 +19,65 @@
...
+function open(organization: Organization | User) {
+  opened.value = organization.id;
+}
+
 onMounted(async () => {
   me.value = await fetchMe();
   if(me.value.organizations.length > 0) {
-    opened.value = me.value.organizations[0].id;
+    open(me.value.organizations[0]);
+    router.replace({name: "organization-datasets", params: {oid: me.value.organizations[0].id}})
```
maudetes, line 79: We shouldn't re-route to first organization dataset if we targeted a specific page right? Ex http://dev.local:7000/fr/beta/admin/#/me

vite.config.dev.ts
```diff
@@ -0,0 +1,75 @@
+import { defineConfig, type UserConfig } from 'vite';
```
maudetes, line 1: (vite.config.dev.ts has been added on master since)

udata_front/theme/gouvfr/assets/js/components/AdminTable/Table/AdminTable.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div class="fr-table">
```
maudetes, line 2: I think we can't scroll on the admin table horizontally in responsive mode.

maudetes (commented):
:star_struck: wow! I like the admin dataset tables a lot!
It seems to work smoothly, and the code seems very readable, well done! :star2:

I think some of my comments are actually more product-oriented.

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
nicolaskempf57: It shows the cell content in a tooltip on smaller screen size

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
nicolaskempf57: This component only uses some informations that are common to both types so both can be provided

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
nicolaskempf57: I think I forgot to reply to your question. 

This is done to follow the "pure presentational" [best practice](https://storybook.js.org/docs/writing-stories/build-pages-with-storybook#pure-presentational-pages) for components or pages of storybook. It moves logic and API request out of individual components to a single wrapper.

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
nicolaskempf57: done !

udata_front/theme/gouvfr/assets/js/components/AdminTable/Table/AdminTable.vue
nicolaskempf57: You're right, this is due to the `display: table` applied to have a full-width table on desktop. I'll have to change this based on the actual breakpoint

udata_front/theme/gouvfr/assets/js/config.ts
nicolaskempf57: soon 👀

udata_front/theme/gouvfr/assets/js/components/AdminTable/Table/AdminTable.vue
nicolaskempf57: The full-width table is now applied on LG breakpoint and wider

udata_front/theme/gouvfr/assets/js/config.ts
maudetes: Could we remove the current metrics fetch logic? And use `dataset.metrics.downloads` that could default to 0 for starting and be replaced by real values when computed?

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+          <QualityScoreTooltip :dataset="dataset"/>
+        </AdminTableTd>
+        <AdminTableTd class="monospace">
+          {{ summarize(dataset.metrics.discussions) }}
+        </AdminTableTd>
+        <AdminTableTd class="monospace">
+          {{ summarize(dataset.metrics.views) }}
+        </AdminTableTd>
+        <AdminTableTd class="monospace">
+          {{ summarize(dataset.metrics.downloads ?? 0) }}
```
maudetes, line 110: `metrics.downloads` is not defined in types

maudetes (approved):
I think we're good! :rocket: Congratulations on this new page! :tada:

package.json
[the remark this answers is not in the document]
nicolaskempf57: Details added to the PR description

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
nicolaskempf57: updated to the correct one and a PR is done to add it to the type from datagouv/components too

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Update translations [#458](https://github.com/datagouv/udata-front/pull/458)
 - use @datagouv/components v1.1.1 [#462](https://github.com/datagouv/udata-front/pull/462)
 
+
```
ThibaudDauce, line 26:
```suggestion
```

udata_front/theme/gouvfr/assets/js/api/datasets.ts
```diff
@@ -42,3 +42,21 @@ export function getFrequenciesUrl () {
...
   return api.get<Array<SpatialGranularity>>(getLocalizedUrl("spatial/granularities/")).then(resp => resp.data);
 }
+
+export type GetOrganizationDatasetsData = {
+  data: Array<Dataset>;
+  next_page: string | null;
+  page: number;
+  page_size: number;
+  total: number;
+};
```
ThibaudDauce, line 52: We don't have a type `Paginator<T>` somewhere?

udata_front/theme/gouvfr/assets/js/components/AdminContentWithTooltip/AdminContentWithTooltip.vue
```diff
@@ -0,0 +1,33 @@
...
+  if(contentRef.value) {
+    tooltipText.value = contentRef.value.textContent || "";
+  }
+});
+</script>
+<style scoped>
+p {
+  line-height: var(--lh) !important;
+}
+</style>
```
ThibaudDauce, line 33: I don't understand this component at all… In the storybook I don't see any tooltip shown? You set the tooltipText from the value of the slot? For me the tooltip text should be a prop? I think I'm lost :-D

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+  (event: 'sort', sort: string): void
+}>();
+
+defineProps<{
+  datasets: Array<Dataset | DatasetV2>
+}>();
+
+const { t } = useI18n();
+
+const sort = ref<SortDirection>('desc');
```
ThibaudDauce, line 148:
```suggestion
const sortDirection = ref<SortDirection>('desc');
```
I think the name `sort` could be ambiguous?

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
ThibaudDauce: I'm not a big fan of signals either… But if it's working why not…

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+  sort.value = direction;
+  sortedBy.value = column;
+  emit('sort', `${sort.value === 'asc' ? "" : "-"}${sortedBy.value}`);
+}
+
+function sorted(column: DatasetSortedBy) {
+  if(sortedBy.value === column) {
+    return sort.value;
+  }
+  return undefined;
```
ThibaudDauce, line 182: I prefer to use `null` when it's a valid response (for me `undefined` is an error in JS, something we forget couldn't be set… `null` is set as `null` for a valid reason, not sure if it's clear :-) )
```suggestion
  return null;
```

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+
+function getFilesCount(dataset: Dataset | DatasetV2) {
+  if(Array.isArray(dataset.resources)) {
+    return dataset.resources.length;
+  }
+  return dataset.resources.total;
+}
+
+function getDatasetLinkToAdmin(dataset: Dataset | DatasetV2) {
+  return admin_root + `dataset/${dataset.id}/`;
```
ThibaudDauce, line 193:
```suggestion
  return `${admin_root}dataset/${dataset.id}/`;
```

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+  }
+}
+
+function getStatusType(dataset: Dataset | DatasetV2): AdminBadgeState {
+  return getStatus(dataset).type;
+}
+
+function getStatusLabel(dataset: Dataset | DatasetV2) {
+  return getStatus(dataset).label;
+}
```
ThibaudDauce, line 214: I this really useful? Could we just call `getStatus()` in the template?

udata_front/theme/gouvfr/assets/js/components/AdminTable/AdminDatasetsTable/AdminDatasetsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+  },
+  private: {
+    label: t("Private"),
+    type: "default",
+  },
+  public: {
+    label: t("Public"),
+    type: "info"
+  }
+}
```
ThibaudDauce, line 170: Instead of having a dict only used once in `getStatus()`, could we inline `return { label: t("Deleted"), type: "error" }` inside the condition of `getStatus()` function?

udata_front/theme/gouvfr/assets/js/components/AdminTable/Table/AdminTableTh.vue
```diff
@@ -0,0 +1,65 @@
...
+        {{ t("Sort") }}
+      </button>
+    </div>
+  </th>
+</template>
+
+<script lang="ts">
+import type { SortDirection } from "../../../types";
+export type AdminTableThProps = {
+  sortable?: boolean;
```
ThibaudDauce, line 26: Do we need this prop? Could we just have a:
```
const sortable = computed(() => props.sorted !== undefined)
``` 

It's only working if `sorted?: SortDirection | null` and we return `null` in my previous comment about not returning `undefined` so we can differentiate between not passing the prop (not sortable) and passing the prop as `null` (sortable but not currently the sort column)

udata_front/theme/gouvfr/assets/js/components/dataset/QualityScore/QualityScoreTooltip/QualityScoreTooltip.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+  <p v-bind="$attrs" :aria-describedby="id" ref="contentRef">
+    <QualityScore :score="dataset.quality.score" />
+  </p>
+  <div class="fr-tooltip fr-placement" :id="id" role="tooltip" aria-hidden="true">
+    <QualityScoreTooltipContent :dataset="dataset" />
```
ThibaudDauce, line 6: The `QualityScoreTooltipContent` is (and will always be?) only used here? Why not inline the logic since the two components are really simple? And why not use the `Tooltip` component here to avoid managing the random ID?

ThibaudDauce (approved):
Everything seems to work, so ok for me. 

My comments are more codestyle so feel free to discard them if it's doesn't match the way you code :-)

udata_front/theme/gouvfr/assets/js/components/AdminContentWithTooltip/AdminContentWithTooltip.vue
nicolaskempf57: You can hover the title column to see of it in action. The goal of the tooltip is to shown content hidden on smaller resolution. The tooltip text is computed to avoid the duplication of the content.

It allows us to call it like this, without passing `dataset.title` as props again.
```
<AdminContentWithTooltip>
    <a class="fr-link" :href="getDatasetLinkToAdmin(dataset)"> 
        {{ dataset.title }}
    </a>
</AdminContentWithTooltip>
```

ThibaudDauce: Capture d’écran 2024-08-26 à 10 40 48

On small screen the title of the dataset is shown and the tooltip show the same thing…

nicolaskempf57: Yes, only if the text is small enough, you can view the other Datasets Table example below to view longer titles

Capture d’écran 2024-08-26 à 10 50 47

ThibaudDauce: Ok, I understand :-) Ok for me then :-)

udata_front/theme/gouvfr/assets/js/api/datasets.ts
nicolaskempf57: I'm taking it from https://github.com/datagouv/udata-front/pull/446

udata_front/theme/gouvfr/assets/js/components/dataset/QualityScore/QualityScoreTooltip/QualityScoreTooltip.vue
nicolaskempf57: I don't recall why I made this one. It seems that Tooltip can be used here too.

Regarding `QualityScoreTooltipContent`, now that the other QualityScore components are in `@datagouv/components`, it should be moved there too to remove duplicates but I think it can be done in a future PR.

nicolaskempf57:
Replaced with https://github.com/datagouv/udata-front/pull/489

PR #372, opened 2024-03-06
Add vite config for dev

nicolaskempf57:
This PR adds a vite config for development only.
It removes the legacy plugin and the bundling of external dependencies.
It also uses a development build of vue instead of the production one.


Jorek57 (approved):

ThibaudDauce (approved):
:1st_place_medal:

PR #381, opened 2024-03-23
Add LOv2 licence to static files

geoffreyaldebert:
- add licence LO2 to static files (instead of etalab blog)


maudetes (commented):
Then we'll need to change URLs pointing to it right?
As in https://www.data.gouv.fr/api/1/datasets/licenses/

maudetes (approved):
We can update links later on

PR #386, opened 2024-04-03
Update vue and fix delimiters error

nicolaskempf57:
Required by https://github.com/etalab/udata-front/pull/385.
The read-more component has a strange bug when imported from a package with an old vue version (same as this udata-front-kit [issue](https://github.com/opendatateam/udata-front-kit/issues/330) and [PR](https://github.com/opendatateam/udata-front-kit/pull/331)).

## Delimiter changes
This PR adds delimiters back to our vue compiler instances because the newer vue version wasn't working without.

New delimiters are `""` (an empty string). 

This works because :
1. vue uses a `typedArray` to store the `charCodes` of the delimiters.
2. In our case, the `typedArray` has a length of 0.
3. When the tokenizer compares its current `charCode` with the delimiter, it uses [`undefined`](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-typedarraygetelement) as the delimiter value.
4. This `undefined` will never match any `charCode` because they are coming from `String.prototype.charCodeAt()` that always return a `number` or `NaN`.

### DIFF
--- a/udata_front/theme/gouvfr/assets/js/index.ts
+++ b/udata_front/theme/gouvfr/assets/js/index.ts
@@ -78,7 +78,7 @@ const configAndMountApp = (el: HTMLElement) => {
   app.component("user-reuse-list", UserReuseList);
 
   // @ts-ignore unset delimiters used in html templates to prevent injections using {{ }}
-  app.config.compilerOptions.delimiters = ["[[[[[", "]]]]]"];
+  app.config.compilerOptions.delimiters = ["", ""];
 
   app.mount(el);
 };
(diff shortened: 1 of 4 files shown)


Jorek57 (approved):
Thanks for the fix !

udata_front/theme/gouvfr/assets/js/index.ts
```diff
@@ -78,7 +78,7 @@ const configAndMountApp = (el: HTMLElement) => {
   app.component("user-reuse-list", UserReuseList);
 
   // @ts-ignore unset delimiters used in html templates to prevent injections using {{ }}
-  app.config.compilerOptions.delimiters = [];
+  app.config.compilerOptions.delimiters = ["[[[[[", "]]]]]"];
```
maudetes, line 81: I'm unsure about this. It may be unlikely to be found in legit user content but still leave room for a dedicated attacker

maudetes (approved):

PR #387, opened 2024-04-03
Fix rename of VisibleDatasetFactory

ThibaudDauce:
Following https://github.com/opendatateam/udata/pull/2997

### DIFF
--- a/udata_front/tests/views/test_dataset.py
+++ b/udata_front/tests/views/test_dataset.py
@@ -6,7 +6,7 @@ from flask import url_for
 
 from udata.core.dataset.factories import (
     ResourceFactory, DatasetFactory, LicenseFactory, CommunityResourceFactory,
-    VisibleDatasetFactory
+    DatasetFactory
 )
 from udata.core.user.factories import UserFactory
 from udata.core.organization.factories import OrganizationFactory
@@ -21,7 +21,7 @@ class DatasetBlueprintTest(GouvfrFrontTestCase):
 
     def test_render_display(self):
         '''It should render the dataset page'''
-        dataset = VisibleDatasetFactory()
+        dataset = DatasetFactory()
         url = url_for('datasets.show', dataset=dataset)
         response = self.get(url)
         self.assert200(response)
--- a/udata_front/tests/views/test_dataset.py
+++ b/udata_front/tests/views/test_dataset.py
@@ -152,6 +152,14 @@ class DatasetBlueprintTest(GouvfrFrontTestCase):
         response = self.get(url_for('datasets.show', dataset=dataset))
         self.assert200(response)
 
+    def test_no_index_on_archived(self):
+        '''It should prevent crawlers from indexing empty datasets'''
+        dataset = DatasetFactory(archived=datetime.utcnow())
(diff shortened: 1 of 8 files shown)


udata_front/tests/views/test_organization.py
```diff
@@ -284,7 +276,6 @@ def test_resources_csv(self):
             for _ in range(3)
         ]
         not_org_dataset = DatasetFactory(resources=[ResourceFactory()])
-        hidden_dataset = DatasetFactory()
```
maudetes, line 287: We now have a `HiddenDatasetFactory` with private=True if you wanna keep tests on hidden datasets?

udata_front/theme/gouvfr/templates/dataset/display.html
```diff
@@ -15,7 +15,7 @@
     'description': dataset.description|mdstrip(60)|forceescape,
     'image': dataset.organization and dataset.organization.logo(external=True) or '',
     'keywords': [_('dataset')] + dataset.tags,
-    'robots': 'noindex, nofollow' if dataset.is_hidden or 'inspire:identifier' in dataset.extras else '',
+    'robots': 'noindex, nofollow' if not dataset.resources or 'inspire:identifier' in dataset.extras else '',
```
maudetes, line 18: We should actually keep on using `is_hidden`, where the logic is.

maudetes (commented):
Thank you for this fix! :pray:
I have minor comments on tests

udata_front/tests/views/test_organization.py
ThibaudDauce: I'm not sure I'm a big fan of `HiddenDatasetFactory`. I start using it but is it really useful to abstract away the fact that it's just `private=True`?

ThibaudDauce: And yes you're right I need to add back the test :-) (didn't understood your comment)

udata_front/theme/gouvfr/templates/dataset/display.html
maudetes: If indexable by data.gouv.fr, it should probably be indexable and used for crawling by search engine?

ThibaudDauce: Done in f70fb40f

udata_front/tests/views/test_organization.py
ThibaudDauce: Done in 5241907a

udata_front/tests/views/test_dataset.py
```diff
@@ -153,14 +152,6 @@ def test_200_if_deleted_but_authorized(self):
         response = self.get(url_for('datasets.show', dataset=dataset))
         self.assert200(response)
 
-    def test_no_index_on_empty(self):
```
maudetes, line 156: Maybe, we could rebrand it to `test_no_index_on_archived`? With an archived dataset as example?

ThibaudDauce: Done in 4fc7cf61

maudetes (approved):
Thank you for the fix! :pray:

nicolaskempf57 (approved):

PR #389, opened 2024-04-09
fix: :bug: add a hasExplore condition to allow display of dataStructure component in ResourceAccordion

Jorek57:
Close https://github.com/etalab/data.gouv.fr/issues/1329


nicolaskempf57 (approved):
Thanks ! Don't forget the changelog

PR #391, opened 2024-04-09
Add CGU button on login page

quaxsze:
fix https://github.com/etalab/data.gouv.fr/issues/1335


maudetes:
We've added a banner to inform users of modality of usage update.

PR #392, opened 2024-04-11
Fix failing captchetat responses

maudetes:
See failing tests in https://app.circleci.com/pipelines/github/etalab/udata-front/3411/workflows/789166fd-c91e-4c3b-82f1-32b508f3882e/jobs/17989.

Indeed, if no Content-Type is set on response, flask_restx expects it to be application/json.
It ends up [wrongfully calling json `dumps()`](https://github.com/python-restx/flask-restx/blob/b38a7d993b3e075feff1bc0af154695b285b6ea2/flask_restx/representations.py#L9) on req.content, which will be bytes.

Even if unexpected, it wasn't failing until then thanks to `ujson`.
`ujson` having been removed since https://github.com/opendatateam/udata/pull/3006/, it now fallbacks on json in [flask_restx make response](https://github.com/python-restx/flask-restx/blob/b38a7d993b3e075feff1bc0af154695b285b6ea2/flask_restx/representations.py#L1).
`json.dumps(b'some bytes')` now fails, on contrary to `ujson.dumps(b'some bytes')`.
We may want to reintroduce ujson for performance reasons, but should probably not apply a json dump on bytes output in any case: https://github.com/ultrajson/ultrajson/issues/264.

To prevent dumping the content, we make sure to build the response explicitly and associate it with upstream captchetat Content-Type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Upgrade vue dependency [#386](https://github.com/etalab/udata-front/pull/386)
 - Add codes and optgroups in Multiselect to display Insee codes and Licence groups [#347] (https://github.com/etalab/udata-front/pull/347)
 - Order Organization's reuses by publishing date [#390](https://github.com/etalab/udata-front/pull/390)
+- Fix failing captchEtat responses [#392](https://github.com/etalab/udata-front/pull/392)
 
 ## 3.5.4 (2024-03-20)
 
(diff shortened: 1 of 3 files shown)


Jorek57 (approved):
Thanks for the fix and the clear explanation about what was happening 🙏

nicolaskempf57 (approved):

maudetes:
cc @ThibaudDauce for information regarding `ujson`, if we wanna reintroduce it.

PR #406, opened 2024-05-02
Add organization type composable

nicolaskempf57:


udata_front/theme/gouvfr/datagouv-components/src/composables/organizations/useOrganizationType.ts
```diff
@@ -0,0 +1,48 @@
...
+    } else if (isLocalAuthority.value) {
+      type.value = LOCAL_AUTHORITY;
+    } else {
+      type.value = COMPANY;
+    }
+  });
+  return {
+    type: readonly(type),
+  };
+}
```
ThibaudDauce, line 48: Why `type` is not a `computed` instead of `watchEffect`?

udata_front/theme/gouvfr/datagouv-components/src/components/Owner/OwnerType.vue
```diff
@@ -0,0 +1,40 @@
...
+      name.value = t("Association");
+      break;
+    case COMPANY:
+      name.value = t('Company');
+      break;
+    default:
+      name.value = t("User");
+      break;
+  }
+});
```
ThibaudDauce, line 39: Why `name` is not a `computed`?

udata_front/theme/gouvfr/datagouv-components/src/components/Owner/OwnerType.vue
```diff
@@ -0,0 +1,40 @@
...
+    case ASSOCIATION:
+      name.value = t("Association");
+      break;
+    case COMPANY:
+      name.value = t('Company');
+      break;
+    default:
+      name.value = t("User");
+      break;
+  }
```
ThibaudDauce, line 38: Maybe put `case 'user'` and add a `default` with a never bloc. In my projects I use this:
```
        default:
            throwOnNever(props.type, `Unknown type ${props.type}`);
```

```
export function throwOnNever(_: never, message: string): never {
    throw new Error(message);
}
```

TypeScript will complain if the first argument is not `never` meaning you forget one variant of the enum.

udata_front/theme/gouvfr/datagouv-components/src/components/Owner/OwnerTypeIcon.vue
```diff
@@ -0,0 +1,19 @@
...
+import { RiBankLine, RiBuilding2Line, RiCommunityLine, RiGovernmentLine, RiUserLine } from "oh-vue-icons/icons";
+import { ASSOCIATION, COMPANY, LOCAL_AUTHORITY, PUBLIC_SERVICE, type OrganizationTypes } from "../../composables/organizations/useOrganizationType";
+import { UserType } from "../../types/users";
+
+addIcons(RiBankLine, RiBuilding2Line, RiCommunityLine, RiGovernmentLine, RiUserLine);
+
+defineProps<{
+  type: OrganizationTypes | UserType;
+}>();
+</script>
```
ThibaudDauce, line 19: You could use a `computed()` name to avoid HTML repetition and use the same `never` trick than above to force the update of this component in case of new variant.

udata_front/theme/gouvfr/datagouv-components/src/composables/organizations/useOrganizationCertified.ts
```diff
@@ -1,22 +1,19 @@
...
     if(!organization) {
       organizationCertified.value = false;
       return;
     }
-    organizationCertified.value = organization.badges.some(badge => badge.kind === PUBLIC_SERVICE) &&
-      organization.badges.some(badge => badge.kind === CERTIFIED);
-  };
-  onMounted(isOrganizationCertified);
+    organizationCertified.value = hasBadge(organization, CERTIFIED) && (isType(organization, PUBLIC_SERVICE) || isType(organization, LOCAL_AUTHORITY));
+  });
```
ThibaudDauce, line 16: This should be a `computed()` for me and not a `watchEffect` but maybe I'm missing something about this project (never use custom `use…` helpers in my Vue apps)

ThibaudDauce (commented):
Just two comments about using `watchEffect` in place of `computed` (but I may miss something about the architecture of the app…) and using TypeScript to warn us on new variant with the `never` trick if you think it's worth it.

nicolaskempf57:
Replaced by https://github.com/datagouv/udata-front/pull/406

PR #412, opened 2024-05-21
Add organization edit to new admin

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1365


Jorek57 (approved):
Good job !

udata_front/theme/gouvfr/assets/js/pages/OrganizationPublishingForm/OrganizationPublishingForm.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-  <div ref="containerRef">
+  <div class="fr-container" ref="containerRef">
+    <Stepper :steps="steps" :currentStep="currentStep"/>
     <Step1CreateOrJoinOrganization
       v-if="currentStep === 0"
       :steps="steps"
```
Jorek57, line 6: Are the steps still necessary if the Stepper is outside of the component ?

nicolaskempf57: No, they are remove from the step components. I'll update the form

udata_front/theme/gouvfr/assets/js/pages/OrganizationPublishingForm/Step2DescribeOrganization.vue
```diff
@@ -275,14 +287,14 @@
...
+    isPublicService: false,
     exists: null as boolean | null
   });
 
   const checkBusinessId = () => {
-    if (organization.business_number_id.length == 0) {
+    if (!organization.business_number_id || organization.business_number_id?.length == 0) {
       return true;
-    } else if (organization.business_number_id.length == 14 && checkOrga.value.exists) {
+    } else if (organization.business_number_id?.length == 14 && checkOrga.value.exists) {
```
ThibaudDauce, line 297: It's weird you need the `?.` because you check for `!organization.business_number_id` at the beginning? A TypeScript bug/limitation?

udata_front/theme/gouvfr/assets/js/pages/OrganizationPublishingForm/Step2DescribeOrganization.vue
```diff
@@ -299,7 +311,7 @@
   const warningRules = {
     acronym: {},
     business_number_id: { custom: checkBusinessId },
-    description: {required, minLengthValue: minLengthWarning(quality_description_length) },
+    description: {required, minLengthValue: minLengthWarning(parseInt(quality_description_length ?? "0")) },
```
ThibaudDauce, line 314: Shouldn't be better to have the `parseInt` in the config to always export `number | null` for `quality_description_length`?

ThibaudDauce (commented):
A few code only comments. Need to check the result on a browser later :-)

udata_front/theme/gouvfr/assets/js/pages/OrganizationPublishingForm/Step2DescribeOrganization.vue
nicolaskempf57: Yes ! Now that's it's a ts file, it's way better to type it in config

udata_front/theme/gouvfr/assets/js/pages/OrganizationPublishingForm/Step2DescribeOrganization.vue
nicolaskempf57: I don't recall, I will try without them. It's because we are still not using @datagouv/component package and the old one doesn't have `business_number_id` in the `Organization` type so the typechecking is not working.

udata_front/theme/gouvfr/assets/js/pages/Admin/Profile/Profile.vue
```diff
@@ -0,0 +1,168 @@
...
+  if(organization.value) {
+    loading.value = true;
+    deleteOrganization(organization.value.id)
+    .then(() => {
+      router.replace("/");
+      window.location.reload();
+    })
+    .catch(() => toast.error(t("An error occured when deleting the organization.")))
+    .finally(() => loading.value = false);
+  }
```
ThibaudDauce, line 152: You could early return here but it's a matter of preference :-)

```suggestion
  if(!organization.value) {
    return;
  }
  loading.value = true;
  deleteOrganization(organization.value.id)
    .then(() => {
      router.replace("/");
      window.location.reload();
    })
    .catch(() => toast.error(t("An error occured when deleting the organization.")))
    .finally(() => loading.value = false);
  }
```

ThibaudDauce (approved):
:+1:

udata_front/theme/gouvfr/assets/js/pages/Admin/Profile/Profile.vue
nicolaskempf57: I tend to use `if` to wrap code when I'm doing it only for the TS type-checker instead of an early exit because the `!organization.value`  can't really exist (the form is hidden without org).

Does this make sense to you ?

PR #414, opened 2024-05-21
Allow quality score warning to be hidden

nicolaskempf57:
This PR updates the non-inline version of the quality component in `@datagouv/components`. It's only used in udata-front-kit. In udata-front, we are only using the jinja version.

This update allows the component to be shown without the quality score warnings. It's required to make the same functionality as our jinja component (see https://github.com/ecolabdata/ecospheres/issues/187).


Jorek57 (approved):

maudetes (approved):
Seems good!

PR #417, opened 2024-05-27
Fix dataset publishing form

nicolaskempf57:
The publishing form was missing a PUT call when uploaded non remote ressources.
See https://github.com/datagouv/data.gouv.fr/issues/1367


Jorek57 (approved):
Good job on the refactoring !

maudetes (approved):
Thank you for the huge refacto! This seems to work :clap:

PR #426, opened 2024-06-07
Fix dataset card link opening another tab

nicolaskempf57:


maudetes (approved):
Thank you! Any other side effect we should check for?

nicolaskempf57:
The version change only touches AppLink.

AppLink and DatasetCard aren't used outside of card-lg in udata-front so there shouldn't be other side effects.

PR #427, opened 2024-06-10
Add chromatic

nicolaskempf57:

PR #428, opened 2024-06-10
Chore/use @datagouv component

nicolaskempf57:
This PR moves udata-front to the new `@datagouv/components` package.

It can wait for https://github.com/datagouv/udata-front/pull/427 to reduce the risk of regression.


nicolaskempf57:
Done in datagouv repo in https://github.com/datagouv/udata-front/pull/439

PR #431, opened 2024-06-12
Add chromatic

nicolaskempf57:
This PR adds Chromatic to udata-front, a subsequent one will add it to datagouv/components too.

To make it work, I had to :
- update to Storybook 8
- update to MSW 2 (required by Storybook 8)
- change some stories due to Breaking Changes in the migration from Storybook 7 to 8 
- fix some stories with errors

Chromatic is configured to do UI Tests (detect changes and run test) and UI Review (show changes from master). One has a development purpose while the Review is more a way for the Product / UI to see, comment and approved / reject visual changes and implementations.

If required later, we can increase the free limits because we are open-source (via chromatic support) or disable UI Review. There is also a `TurboSnap` that we can use after more build and PR to reduce the number of snapshot taken.


udata_front/theme/gouvfr/assets/js/components/MarkdownEditor/ImageModal/ImageModal.vue
```diff
@@ -128,7 +128,7 @@ function close() {
 function send() {
   validateRequiredRules().then(valid => {
     if(valid) {
-      emit('send', {...form});
+      emit('sendModal', {...form});
```
ThibaudDauce, line 131: Not sure why this change but ok :-) `ImageModalButton` keep the `send` event?

ThibaudDauce (approved):
:+1:

PR #437, opened 2024-07-08
Add organization type composable

nicolaskempf57:
Replaces https://github.com/datagouv/udata-front/pull/406
First part of https://github.com/datagouv/data.gouv.fr/issues/1352


udata_front/theme/gouvfr/datagouv-components/src/composables/organizations/useOrganizationCertified.ts
```diff
@@ -1,22 +1,17 @@
...
-      return;
-    }
-    organizationCertified.value = organization.badges.some(badge => badge.kind === PUBLIC_SERVICE) &&
-      organization.badges.some(badge => badge.kind === CERTIFIED);
-  };
-  onMounted(isOrganizationCertified);
+      return false;
+    };
+    return hasBadge(organization, CERTIFIED) && (isType(organization, PUBLIC_SERVICE) || isType(organization, LOCAL_AUTHORITY));
+  });
```
ThibaudDauce, line 14: I think this should be an helper `isOrganizationCertified(org: Organization)` and the `computed` should be in the component:
```
const isCertified = computed(() => isOrganizationCertified(org.value))
```

udata_front/theme/gouvfr/datagouv-components/src/composables/organizations/useOrganizationType.ts
```diff
@@ -0,0 +1,50 @@
...
+    } else if (isLocalAuthority.value) {
+      return LOCAL_AUTHORITY;
+    } else {
+      return COMPANY;
+    }
+  });
+  return {
+    type,
+  };
+}
```
ThibaudDauce, line 50: Same, I think it should be an helper `getOrganizationType(org: Organization)`  and the computed moved to the component to avoid leaking Vue concepts:
```
const organizationType = computed(() => getOrganizationType(org.value))
```

ThibaudDauce (approved):
Ok for the PR (but I still think we leak too much VueJS outside components :-) )

Jorek57 (approved):

PR #438, opened 2024-07-09
Add orga selector to dataset publishing

Jorek57:


nicolaskempf57 (approved):

udata_front/theme/gouvfr/assets/js/pages/PublishingForm/Step2DescribeDataset.stories.ts
```diff
@@ -111,7 +147,7 @@ const args: Step2DescribeDatasetProps = {
   ]
 };
 
-export const Step2: StoryObj<typeof meta> = {
+export const Step2WithNoOrganization: StoryObj<typeof meta> = {
```
nicolaskempf57, line 150: Unrelated to your changes, somehow Chromatic has a hard time dealing with `waitFor` in the case of form validation. You can replace waitFor [with a timeout to make it work](https://github.com/datagouv/udata-front/pull/431/files#diff-995097d62d0ae471b325dd97267e3dcf145dc7ba352334c6b325397a23a65f4eR40).

nicolaskempf57: After another look at this, it seems that Chromatic might be right, this case should be checked to see if we can send the form without an organization

PR #452, opened 2024-07-23
Update deps following udata dependencies update

maudetes:
See also https://github.com/datagouv/udata-front/pull/448.

Applied pip-compile to all dependencies.


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/jazzband/pip-tools
-    rev: 6.2.0
+    rev: 7.4.1
```
maudetes, line 3: Needed following https://github.com/jazzband/pip-tools/issues/1617.

PR #453, opened 2024-07-23
Fix tests for oembed cors

ThibaudDauce:
Following https://github.com/opendatateam/udata/pull/3074 we need to send an `Origin` header to get the `Access-Control-Allow-Origin` otherwise the request is not a cross origin request.

### DIFF
--- a/udata_front/tests/views/test_oembed.py
+++ b/udata_front/tests/views/test_oembed.py
@@ -31,7 +31,7 @@ class OEmbedAPITest:
 
         url = url_for('api.oembed', url=dataset.external_url)
 
-        response = api.get(url)
+        response = api.get(url, headers={ 'Origin': 'http://localhost' })
         assert200(response)
         assert_cors(response)
         assert 'html' in response.json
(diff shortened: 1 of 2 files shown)


udata_front/tests/views/test_oembed.py
```diff
@@ -26,139 +30,164 @@ class OEmbedAPITest:
...
-        assert 'html' in response.json
-        assert 'width' in response.json
-        assert 'maxwidth' in response.json
-        assert 'height' in response.json
-        assert 'maxheight' in response.json
-        assert response.json['type'] == 'rich'
-        assert response.json['version'] == '1.0'
-        card = theme.render('dataset/card-xs.html', dataset=dataset)
-        assert card in response.json['html']
+        assert "html" in response.json
```
maudetes, line 85: I would not modify quotes in this PR since single quotes are still the norm.
We should probably wait for a global ruff pass?

maudetes (approved):
Do you think there could be actual legit usages of oembed without a Origin header?

ThibaudDauce:
> Do you think there could be actual legit usages of oembed without a Origin header?

I don't think so. If a browser need a cross origin verification it will send an `Origin` header. https://fetch.spec.whatwg.org/#cors-request

udata_front/tests/views/test_oembed.py
ThibaudDauce: Done :-)

PR #456, opened 2024-07-24
Add captchetat to reset route

ThibaudDauce:
Code from https://github.com/datagouv/udata-front/pull/273.

Moved the captcha check before the `super().validate()` to prevent checking email existance without captcha.

Fix https://github.com/datagouv/data.gouv.fr/issues/1403


nicolaskempf57 (approved):
Nice 🙏🏻

PR #458, opened 2024-07-29
Fix/components i18n

nicolaskempf57:
This PR adds missing translations and update the `vue-i18n-extract` config file now that udata-front-plugins-helper doesn't contain any vue files.


maudetes (approved):
:pray: Thank you!

PR #466, opened 2024-07-30
Update dsfr for datagouv/components

nicolaskempf57:
This PR should be tested with udata-front using the same version of gouvfr/dsfr (e.g. https://github.com/datagouv/udata-front/pull/371) otherwise NPM magic will use the udata-front version and not the updated one we want here.


ThibaudDauce (approved):

nicolaskempf57:
> Do you have this new "bottom border" too? 

This really looks like a DSFR things, I'll look 🙏🏻

PR #468, opened 2024-07-31
Fix resource translation

nicolaskempf57:
This PR fixes the translation `Use our tool to get an overview of data, learn about different columns or perform filters and sorts.` not found.

It upgrades the `vue-i18n` to fix the regression (see https://github.com/intlify/vue-i18n/issues/1717 for reference).

There is also a slight visual fix to prevent the `explore data` button to show its content on 2 lines.

### DIFF
--- a/udata_front/theme/gouvfr/datagouv-components/CHANGELOG.md
+++ b/udata_front/theme/gouvfr/datagouv-components/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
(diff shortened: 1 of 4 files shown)


maudetes:
I think we have a similar issue on report modal, can you check that it fixes it please? :pray:

maudetes (approved):
Thank you for the fix!

magopian (approved):
LGTM! Must have been a nightmare to hunt down this issue 😬

PR #471, opened 2024-08-01
Fix translations

nicolaskempf57:
Fixes reports translation, upgrades vue-i18n and use newest datagouv/component release.

See [components release v1.1.2](https://github.com/datagouv/udata-front/releases/tag/%40datagouv%2Fcomponents%401.1.2) for more details.

### DIFF
--- a/udata_front/theme/gouvfr/assets/js/components/Report/ReportModal.vue
+++ b/udata_front/theme/gouvfr/assets/js/components/Report/ReportModal.vue
@@ -74,7 +74,7 @@
                     :hasError="fieldHasError('message')"
                     :errorText="getErrorText('message')"
                     v-model="form.message"
-                    :placeholder="t('Reason of your report.\nDon\'t include any personal data.')"
+                    :placeholder="t(`Reason of your report.{newline}Don't include any personal data.`)"
                   />
                 </div>
                 <div class="fr-modal__footer">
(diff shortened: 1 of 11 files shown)


udata_front/theme/gouvfr/assets/js/components/Report/ReportModal.vue
```diff
@@ -74,7 +74,7 @@
                     :hasError="fieldHasError('message')"
                     :errorText="getErrorText('message')"
                     v-model="form.message"
-                    :placeholder="t('Reason of your report.\nDon\'t include any personal data.')"
```
maudetes, line 77: The issue was with the quote character right? not the newline?

nicolaskempf57: The issue was the use of `\` in the translation key. We can keep the `\n` on the value side.

udata_front/theme/gouvfr/assets/js/locales/en.json
```diff
@@ -449,7 +449,7 @@
   "Please only report in case of serious concern.": "Please only report in case of serious concern.",
   "See our usage policy": "See our usage policy",
   "Report reason": "Report reason",
-  "Reason of your report.\\nDon\\'t include any personal data.": "Reason of your report.\\nDon\\'t include any personal data.",
+  "Reason of your report.{newline}Don't include any personal data.": "Reason of your report.\nDon't include any personal data.",
```
maudetes, line 452: I'm not sure why, but when running `inv i18n` again, it removes the value? Should we recompile these with Crowdin?

nicolaskempf57: I think this was the case because I didn't update all the files. It's working now that all files have the same keys.

maudetes (approved):
Perfect! :ok_hand: 

Maybe add an entry in the udata-front changelog (regarding translations and components upgrade)

PR #472, opened 2024-08-13
Show organization type

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1352
Closes https://github.com/datagouv/data.gouv.fr/issues/1277

This PR adds organisation type to the producer metadata.


udata_front/theme/gouvfr/assets/js/config.ts
```diff
@@ -288,7 +288,7 @@ export const search_autocomplete_debounce = _jsonMeta("search-autocomplete-debou
 
 export const tabular_api_url = _meta("tabular-api-url");
 
-export const tabular_page_size = _meta("tabular-page-size");
+export const tabular_page_size = parseInt(_meta("tabular-page-size") ?? "10");
```
maudetes, line 291: Why can't it be left to null if not defined?

udata_front/frontend/helpers.py
```diff
@@ -211,6 +218,18 @@ def owner_name_acronym(obj):
...
+@front.app_template_global()
+def organization_type(org: Organization):
+    if org.public_service:
+        return PUBLIC_SERVICE
+    elif org.local_authority:
+        return LOCAL_AUTHORITY
+    elif org.association:
+        return ASSOCIATION
+    else:
+        return COMPANY
```
maudetes, line 230: I think the issue with this approach is that a new organization without badge will be shown as `company` (not only the badge but the wording on its page) by default, which is wrong.

udata_front/frontend/helpers.py
```diff
@@ -211,6 +218,18 @@ def owner_name_acronym(obj):
     return ''
 
 
+@front.app_template_global()
+def organization_type(org: Organization):
+    if org.public_service:
```
maudetes, line 223: Should local authority be prioritized over public service when showing logo and having both badges? cc @agarrone

udata_front/theme/gouvfr/templates/dataset/display.html
maudetes, line 1: I'm not sure, but don't we want to add it on jinja dataset or organization cards for example?

maudetes (commented):
Thank you for this PR!

I may be reacting off beat, but the icon seems larger than the associated text which is not the case on the Figma?

nicolaskempf57:
> Thank you for this PR!
> 
> I may be reacting off beat, but the icon seems larger than the associated text which is not the case on the Figma?

Yes, the design was using 10x10px icons. Instead, we are using a scale size of 1.2 (so around 19px). 
Using smaller icons, even 16px, was making them really hard to understand. The line-height of the text is 21px, so changing the icon size doesn't change the overall design.

udata_front/frontend/helpers.py
nicolaskempf57: My understanding was that a local authority shouldn't have public service in their badge array. 
Also, we can change the condition but this one is only a "translation" of the existing one in datagouv/components so a new release will be necessary to have the change everywhere.

udata_front/theme/gouvfr/assets/js/config.ts
nicolaskempf57: It is always defined in `udata_tabular_preview`, but Typescript doesn't know it and we can't parse a `string | null` to `int`. I will changed the default to match the one from `udata_tabular_preview` and datagouv/components.

udata_front/frontend/helpers.py
nicolaskempf57: I didn't realise we wanted a case without any type. If we do, we might want to make a new datagouv/components before this PR to change the condition [there too](https://github.com/datagouv/udata-front/blob/feat/show-organization-type/udata_front/theme/gouvfr/datagouv-components/src/composables/organizations/useOrganizationType.ts#L44).

nicolaskempf57: TS changes done in https://github.com/datagouv/udata-front/pull/475

udata_front/frontend/helpers.py
nicolaskempf57: TS change done in https://github.com/datagouv/udata-front/pull/475 too

udata_front/theme/gouvfr/templates/dataset/display.html
nicolaskempf57: We can make a new macro that use the `<OrganizationNameWithCertificate>` component for the homepage dataset and the organization cards before their update.

We can't make a complete non-js version without custom logic because some icons are coming from RemixIcon and not the DSFR.

nicolaskempf57: This is done by only using the `<owner-type-icon>` js component. So in the event of a non-js environment, the rest will be available.

udata_front/frontend/helpers.py
maudetes: I know we'll treat the root cause in https://github.com/datagouv/data.gouv.fr/issues/1433 but I think the order should match the one in datagouv/components `useOrganizationType`? Else, we may have discrepancy before applying a migration as in http://dev.local:7000/fr/organizations/grand-chambery/#/datasets.

maudetes (commented):
I think we're mostly clear!

> Yes, the design was using 10x10px icons. Instead, we are using a scale size of 1.2 (so around 19px).
Using smaller icons, even 16px, was making them really hard to understand. The line-height of the text is 21px, so changing the icon size doesn't change the overall design.

Upon discussion with @agarrone, I think we could like to give it a go with 16px icons?

maudetes (approved):
Thank you again for this PR!

Let's deal with the icon size question in a follow-up PR :)

PR #474, opened 2024-08-16
Add new dataset metric

nicolaskempf57:
Add new dataset metric to Dataset type

Follows https://github.com/opendatateam/udata/pull/3042

Required for https://github.com/datagouv/udata-front/pull/371


maudetes (approved):
:ok_hand:

PR #475, opened 2024-08-19
Improve organization type logic

nicolaskempf57:
This PR changes the detection order of type to show `local authority` even for legacy cases with both `public service` and `local authority`.

It also adds a new `other` type without icon that shouldn't be shown on the organization page.


ThibaudDauce (approved):
:+1:

PR #485, opened 2024-08-26
Add ellipsis for long org name on dataset card

ThibaudDauce:
Fix https://github.com/datagouv/data.gouv.fr/issues/1429

### DIFF
--- a/udata_front/theme/gouvfr/assets/less/content/typography.less
+++ b/udata_front/theme/gouvfr/assets/less/content/typography.less
@@ -314,7 +314,7 @@ each(@breakpoints, .(@bk, @_k, @_i) {
     margin-right: 0.25rem;
 }
 
-.dash-after-sm::after {
+.dash-after-sm::before, .dash-before-sm::after {
     content: "—";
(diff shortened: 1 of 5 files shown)


udata_front/theme/gouvfr/assets/less/content/typography.less
```diff
@@ -318,4 +322,15 @@ each(@breakpoints, .(@bk, @_k, @_i) {
     margin-bottom: auto;
     color: @grey-380;
     font-size: 0.6rem;
-  }
+}
+
+.dash-before-sm::before {
```
nicolaskempf57, line 327: Can you add the class to the rule above ? It should work with `.dash-after-sm::after, .dash-before-sm::before` no ? Also, I don't know why I didn't do it for `.dash-before::before` and `.dash-after::after`.

ThibaudDauce: Fixed in 4fc79013

nicolaskempf57 (approved):

PR #493, opened 2024-08-29
Fix dataset card visual glitches

nicolaskempf57:
Fixes a tooltip bug and a button bug reported bug by @agarrone 

Changes the logic for ellipsis for long organizaiton name from CSS to vue-clamp

This introduces a new package required for https://github.com/datagouv/data.gouv.fr/issues/1422


ThibaudDauce (approved):
Seems ok for me. 

I prefer to use `<TextClamp />` instead of `<text-clamp/>` to distingate between native components and VueJS components but not a big deal.

PR #498, opened 2024-09-03
Update sentry metadata

nicolaskempf57:
1/2 of https://github.com/datagouv/data.gouv.fr/issues/1438


ThibaudDauce (approved):

maudetes:
Later on, do we want to use a specific `SENTRY_UDATA_FRONT_PUBLIC_DSN` to separate with backend one?
We could also use the same project but with some tag to easily differentiate.

PR #501, opened 2024-09-03
Use new reuse card

nicolaskempf57:
Use new reuse card 

Part of https://github.com/datagouv/data.gouv.fr/issues/1442


udata_front/theme/gouvfr/templates/reuse/oembed.html
nicolaskempf57, line 1: Jinja content is moved to the oembed file to prevent them from failing. A better long term solution should be worked on in the future

ThibaudDauce (approved):
:+1:

PR #504, opened 2024-09-05
Disconnect from proconnect on logout

magopian:
Fixes [#1454](https://github.com/datagouv/data.gouv.fr/issues/1454)

⚠️  this PR is based on [#482]


udata_front/theme/gouvfr/templates/header.html
```diff
@@ -92,10 +92,17 @@
                                     </a>
                                 </li>
                                 <li>
-                                    <a
-                                        href="{{ url_for('security.logout') }}"
-                                        class="fr-btn fr-icon-logout-box-r-line"
-                                    >
+                                    {% if current_user.is_authenticated and current_user.password %}
```
maudetes, line 95: I think the `password` is not a good discriminate for a ProConnect session since a user can both use datagouv login and ProConnect.

udata_front/views/proconnect.py
```diff
@@ -47,3 +53,45 @@ def auth():
...
+
+@blueprint.route('/logout_oauth', endpoint='logout_oauth')
+def logout():
+    # At the time of this writing, authlib didn't implement OpenIDC session management:
+    # https://github.com/lepture/authlib/issues/292
+    # So we implement it ourselves. This code may be simplified (or even removed?) in the future
+    # if we update to a version that supports it.
+    metadata = oauth.proconnect.load_server_metadata()
+    end_session_endpoint = metadata["end_session_endpoint"]
+    # Generate a random state that we send to ProConncet, they'll return it so we can check it.
```
maudetes, line 66:
```suggestion
    # Generate a random state that we send to ProConnect, they'll return it so we can check it.
```

maudetes (approved):
Thank you! Seems to work as expected! :)

udata_front/theme/gouvfr/templates/header.html
magopian: ah yes of course, we talked about that, and in the meantime I changed the code to [store the `id_token` in the session](https://github.com/datagouv/udata-front/pull/504/files#diff-6d88bad001db215ea314bfaf930dfc0a219958455a2c8d0241e3dc47be001a4dR27), so this could be a good check, right?

udata_front/views/proconnect.py
magopian: Good catch, thanks 🙏

udata_front/theme/gouvfr/templates/header.html
maudetes: I think so indeed, thank you!

PR #505, opened 2024-09-05
Add some information on how to setup and run the cypress e2e tests

magopian:


README.md
```diff
@@ -266,6 +266,24 @@ Writing tests is very easy thanks to its syntax :
   });
 ```
 
+To run the Cypress tests, you need a local webserver. Run the following command in the same Python virtual environment as `udata`:
```
nicolaskempf57, line 269: I think the command lines should be in the opposite order : the cfg copy first, then the prepare and finally the serve

magopian: Ah, yes, you're right, fixing this right away 🙏

nicolaskempf57 (approved):
Looks good to me

PR #515, opened 2024-09-13
Use datagouv/components without build

nicolaskempf57:
`@datagouv/components` now exports 3 files :
- `@datagouv/components` the js bundled
- `@datagouv/components/dist/style.css` the css bundled
- `@datagouv/components/ts` the source `index.ts`

udata-front is now using the source directly instead of relying on the build version.
This doesn't change anything for the end-user, everything is bundled during udata-front build.

Most of the changes are imports moved from `@datagouv/components` to `@datagouv/components/ts`.

I moved 2 JS files to TS because it was easier to do so than to try to see if vite handle TS import in JS file.

I know that `useUserAvatar` should be moved to a `helper` but I didn't want to clutter this PR.


udata_front/theme/gouvfr/assets/js/helpers.ts
nicolaskempf57, line 1: Removed imports here were unused and the `_` is added to follow typescript convention with unused variable

udata_front/theme/gouvfr/assets/less/style.less
nicolaskempf57, line 1: Compiled css isn't necessary anymore and both the components and udata-front CSS is generated with vite in a single file

udata_front/theme/gouvfr/datagouv-components/src/components/ReuseCard/ReuseCard.vue
nicolaskempf57, line 1: unused import

udata_front/theme/gouvfr/datagouv-components/src/composables/useComponentsForHook.ts
nicolaskempf57, line 1: This composable was used when the preview was outside of the resource and this is not the case anymore

udata_front/theme/gouvfr/datagouv-components/src/helpers/i18n.ts
```diff
@@ -1,11 +1,10 @@
 import dayjs from "dayjs/esm";
 import LocalizedFormat from 'dayjs/esm/plugin/localizedFormat';
 import RelativeTime from "dayjs/esm/plugin/relativeTime";
-import { DefineLocaleMessage, I18n, Locale, createI18n } from "vue-i18n";
+import { type I18n, createI18n } from "vue-i18n";
 import messages from '@intlify/unplugin-vue-i18n/messages';
 import { config } from "../config";
 import { api_root_absolute } from "../api/api";
-import { getRegisteredTranslations } from "@etalab/udata-front-plugins-helper";
```
nicolaskempf57, line 8: everything related to udata-front-plugins-helper is unused so it's removed.

removing udata-front-plugins-helper is the next step but it significant enough to be a separate PR

vite.config.dev.ts
```diff
@@ -43,8 +46,10 @@ async function getConfig(): Promise<UserConfig> {
         hook: 'writeBundle'
       }),
     ],
+    envDir: "udata_front/theme/gouvfr/datagouv-components",
     build: {
       minify: false,
+      cssCodeSplit: false,
```
nicolaskempf57, line 52: We don't want css splitting because we are doing the import ourself without manifest

vite.config.dev.ts
```diff
@@ -43,8 +46,10 @@ async function getConfig(): Promise<UserConfig> {
         hook: 'writeBundle'
       }),
     ],
+    envDir: "udata_front/theme/gouvfr/datagouv-components",
```
nicolaskempf57, line 49: We are using the `.env` file from the components folder otherwise the components source are missing expected variables

udata_front/theme/gouvfr/datagouv-components/src/helpers/index.ts
```diff
@@ -39,14 +39,14 @@ export const filesize = (val: number) => {
   return `${val.toFixed(1)}Y${suffix}`
 }
 
-export const summarize = (val: number) => {
+export const summarize = (val: number, fractionDigits = 0) => {
```
nicolaskempf57, line 42: This was the test I did to check that the modification is showing on udata-front

udata_front/theme/gouvfr/datagouv-components/package.json
```diff
@@ -7,10 +7,22 @@
...
+      "import": "./dist/data-gouv-fr-components.js"
+    },
+    "./ts": "./src/index.ts",
+    "./dist/style.css": "./dist/style.css"
+  },
   "module": "./dist/data-gouv-fr-components.js",
   "types": "./dist/data-gouv-fr-components.d.ts",
   "scripts": {
     "dev": "vite",
+    "dev:udata_front": "vite build --watch",
```
maudetes, line 25: What is this for?

maudetes (commented):
Thank you a lot for the clean PR explanation! I am reviewing this PR early for once :stuck_out_tongue: 
Indeed, I think we can start with this one before adding more refactoring.

I have some questions as I stumble on this:
* Other datagouv-components users should update their import path as well?
* I am unsure of the datagouv components release process now?

I think we can update udata-front & datagouv-components Readme files.

udata_front/theme/gouvfr/datagouv-components/package.json
nicolaskempf57: Oops, this is a leftover of a previous trial

ThibaudDauce (approved):
:100:

maudetes (approved):
:ok_hand:

PR #527, opened 2024-09-18
Add reuses page to the new administration

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1484


udata_front/theme/gouvfr/assets/js/components/AdminPublishButton/AdminPublishButton.vue
```diff
@@ -1,13 +1,28 @@
...
+const link = getLink();
+
+function getLink() {
+  switch(props.type) {
+    case 'dataset':
+      return dataset_publishing_form_url;
+    case 'reuse':
+      return reuse_publishing_form_url
+  }
+}
```
ThibaudDauce, line 27: Need to be a `computed` (even if there is no reason for the `type` prop to change after mounting the component :-D )

ThibaudDauce (approved):
Nice!

PR #531, opened 2024-09-19
Fix resource scroll

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1483


maudetes (approved):
It seems to work, thank you for your fix!

PR #541, opened 2024-09-23
Fix translations

nicolaskempf57:
Translations with `\` in their key aren't picked up by vue-i18n.


maudetes (approved):
:pray: thank you for the fix!

It's sad that it fails silently though, because it may occur again in the future

PR #548, opened 2024-09-30
Add special style to tabular API dataservice card

ThibaudDauce:
Related to https://github.com/datagouv/data.gouv.fr/issues/1467


nicolaskempf57 (approved):
We should change the `VITE_TABULAR_API_DATASERVICE_ID` to the real id when we have it

ThibaudDauce:
> We should change the `VITE_TABULAR_API_DATASERVICE_ID` to the real id when we have it

The `.env` file is used only in tests? We set the env variable somewhere else for production or not?

maudetes:
I may be missing something but I think we would not override `config.tabular_api_dataservice_id` with the `TABULAR_API_DATASERVICE_ID` defined in [our udata settings](https://github.com/opendatateam/udata/blob/e40b20306c4f89407f118a3326f1fcead9fdf539/udata/settings.py#L543) (and set in udata.cfg)?
I've added https://github.com/datagouv/udata-front/pull/548/commits/7114840600f929aad451aa5e189bd43fcd26fd21 to use the udata setting value.

PR #549, opened 2024-10-01
Add user datasets page

ThibaudDauce:


nicolaskempf57 (commented):
The "download catalog" button is still using the wrong URL.

Edit : I don't know if we want to show it for a user

udata_front/theme/gouvfr/assets/js/api/datasets.ts
```diff
@@ -52,6 +52,13 @@ export async function getOrganizationDatasets(oid: string, q: string, page: numb
   return resp.data;
 }
 
+export async function getUserDatasets(userId: string, q: string, page: number, pageSize: number, sortDirection: string) {
+  const resp = await api.get<PaginatedArray<Dataset>>(getLocalizedUrl(`datasets/`), {
+    params: { q, sort: sortDirection, page_size: pageSize, page, owner: userId }
```
maudetes, line 57: I don't think you get your private, deleted & archived datasets on this route sadly :cry:

PR #557, opened 2024-10-10
Fix beta admin

nicolaskempf57:
See https://github.com/datagouv/data.gouv.fr/issues/1521

Everything is done except the new header (to do in in an upcoming PR) and the responsive (not on Figma yet).

### DIFF
--- a/udata_front/theme/gouvfr/assets/js/pages/Admin/Members/Members.vue
+++ b/udata_front/theme/gouvfr/assets/js/pages/Admin/Members/Members.vue
@@ -70,10 +70,9 @@
                 :height="12"
                 :width="12"
                 name="ri-mail-line"
-                class="fr-mr-1v"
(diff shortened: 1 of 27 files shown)


udata_front/theme/gouvfr/assets/js/pages/Admin/Members/Members.vue
```diff
@@ -45,49 +45,57 @@
...
+                name="ri-mail-line"
+                class="fr-mr-1v"
               />
-            </td>
-          </tr>
-        </tbody>
-      </AdminTable>
-    </div>
+              <Tooltip>
+                <TextClamp :text="member.user.email" :auto-resize="true" :max-lines="1" />
```
maudetes, line 76: I think we can now have the member email first letter slightly cropped
image

PR #562, opened 2024-10-15
 low enphasis last element of chart

ThibaudDauce:


nicolaskempf57 (commented):
I think StatBox should be updated to call SmallChart with the new `lastWithLowEnphasis` props

nicolaskempf57 (approved):

PR #567, opened 2024-10-18
Creates reuses as private during the publishing form and fixes buttons

nicolaskempf57:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1531


maudetes (approved):
Thank you for the fix! :ok_hand:

PR #577, opened 2024-11-04
Migrate organization badges label to lowercase

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/1515 on front side
Should be in sync with https://github.com/opendatateam/udata/pull/3182

I guess we should use `api/1/organizations/badges` instead to get the badges, but we'll need to know which badges are orga types (certified is not for example).
I didn't tackle this yet!


magopian (approved):
yep! Thanks for the last piece in the puzzle of updating the badges 🙏 
🚢

PR #579, opened 2024-11-05
Add organization type filter to dataset search

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1431
Based on https://www.figma.com/design/7BmOEaby2vSJ88zuBURpHy/%F0%9F%8E%B1-Typologie-des-organisations?node-id=0-1&node-type=canvas&t=f6U99lyL79qnBIGY-0

2 types from Figma aren't shown in the actual code :
- other : this is not actually possible to filter datasets from organization without any badge
- user : we can't filter datasets without organization

But, we have a "all type" (the default value) that was missing from the Figma design.

We can implement the missing filters but I don't think they are meaningful for our users. Also, if users really want to filter for dataset onwed by other users, we should put the option outside of "organization type".


maudetes (approved):
Thank you for this PR! :rocket:

PR #591, opened 2024-11-08
Fix dataservice pages

nicolaskempf57:


maudetes (approved):
Thank you for the fix!

PR #596, opened 2024-11-15
Rename wrongly named function

nicolaskempf57:
We have an helper function called like a composable, but it's not a composable.

Now this is fixed !


maudetes (approved):
Thank you for the fix!

PR #601, opened 2024-11-18
Add Dataservice search in org page

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1570


udata_front/views/organization.py
```diff
@@ -86,6 +86,9 @@ def get_context(self):
         if self.organization.deleted and not can_view.can():
             abort(410)
 
+        dataservices = Dataservice.objects(
```
maudetes, line 89: I think you need to filter these out if not visible.
Since we're using the search service, I guess we won't be able to display these if the user is allowed.

maudetes (approved):
:rocket:

PR #610, opened 2024-11-26
Disable session tracking

nicolaskempf57:
Sentry is sending so many session requests that we want to disable it for now.

It could be a problem with how many sentry instance we have on a given page, so maybe we can try to have [a single one with multiple vue apps](https://docs.sentry.io/platforms/javascript/guides/vue/features/multiple-apps/) later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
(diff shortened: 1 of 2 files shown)


maudetes (approved):
:pray: could you test locally?

nicolaskempf57:
> 🙏 could you test locally?

Done right now ! From 28 requests down to 0 🚀

PR #621, opened 2024-12-12
Use beta admin as default

nicolaskempf57:
See https://github.com/datagouv/front-end/pull/207 for the same on front-end side


maudetes (approved):
:rocket: :clap:

Do we want to modify all the other links pointing to the administration (edit button, etc.) as well?

udata_front/frontend/helpers.py
```diff
@@ -109,6 +109,20 @@ def in_url(*args, **kwargs):
     )
 
 
+@front.app_template_global()
+def beta_admin_url_for(path, fallback_path, **kwargs):
```
maudetes, line 113: We're missing some routes. They have been documented in https://github.com/datagouv/data.gouv.fr/issues/1648 and can be added in future iterations.

Membership Request notification @ https://github.com/datagouv/udata-front/blob/7cc5df3e8aa3e7af80aaa36877f81c2dbad67801/udata_front/theme/gouvfr/templates/mail/membership_request.html#L41 are actually not used. I think udata's templates are used and I guess we can't rely on beta_admin_url_for there. Let's ignore it for now?

nicolaskempf57: I agree, I think we'll have to work on the mailing system at some point because it was not thought with this usecase in mind.

PR #622, opened 2024-12-13
Add dataservice search to header

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1584


maudetes (approved):
Looks good, thank you!

As an additional question, I'm confused about keyboard navigation on this component? Weren't we able to use the arrow key to access the different object search? I could not make it work in production ether.

nicolaskempf57:
> As an additional question, I'm confused about keyboard navigation on this component? Weren't we able to use the arrow key to access the different object search? I could not make it work in production ether.

Indeed, keyboard navigation seems to be broken right now. I'll make a new issue for this

nicolaskempf57:
Keyboard issue : https://github.com/datagouv/data.gouv.fr/issues/1640

PR #633, opened 2025-01-10
fix: BaseDataset.temporal_coverage type

abulte:


udata_front/theme/gouvfr/datagouv-components/src/types/datasets.ts
```diff
@@ -25,7 +25,10 @@ export type BaseDataset = Owned & {
   tags: Array<string> | null;
   license: string;
   frequency: string;
-  temporal_coverage: string;
+  temporal_coverage: {
+    start: string;
+    end: string;
+  } | null;
   frequency_date: Date | null;
```
abulte, line 32: I think this should be `string | null` BTW.

nicolaskempf57: You're right, the existing types were the ones from the publishing form but not the ones from the AP. We can change them to better fit their current use.

nicolaskempf57 (approved):

nicolaskempf57:
> Tests are failing on Chromatic:
> 
> > Error: ✖ Missing project token
> 
> It might be because I'm doing a PR from a fork? This does not look related to the changes I made.

You're right, secrets are not shared with forks for security reason. It doesn't really matter here

nicolaskempf57:
> Tests are failing on Chromatic:
> 
> > Error: ✖ Missing project token
> 
> It might be because I'm doing a PR from a fork? This does not look related to the changes I made.

I'll merge without the chromatic steps when the others are finished.

PR #634, opened 2025-01-10
New newsletter design

nicolaskempf57:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1633


maudetes (commented):
Thank you for this new footer version :sparkles: 

Things I see that differ between udata-front and front-end:
- [x] logo twitter
- [ ] `Open data` h3 size is not similar (due to datagouv components override?)
- [ ] `Follow us on social media` is not vertically aligned with `Subscribe to our newsletter`
- [ ] `See our next events` button is blue or black
- [ ] data.gouv.fr logo size
- [ ] `Data catalog` does not exist in front-end
- [ ] `A product of the Direction Interministérielle du Numérique(DINUM).` horizontal alignment differs between the two (maybe due to language picker missing?)
- [ ] `Licenses` is to be preferred instead of `Licences`
- [ ] horizontal spacing between Licenses and Terms of use

nicolaskempf57:
> * [ ]  `Data catalog` does not exist in front-end

I added a `csvDatasetId` runtime variable to pass the id, to match the udata-front side

> * [ ]  `A product of the Direction Interministérielle du Numérique(DINUM).` horizontal alignment differs between the two (maybe due to language picker missing?)

Indeed, this is the cause of this misalignment

maudetes (approved):
I think we're good!

PR #650, opened 2025-01-24
Fix security tests after udata changes

nicolaskempf57:
Udata adds trailing slashes to security URLs but we are checking the old ones here


maudetes (approved):
:pray:

Issue #668, opened 2025-03-04
InformationPanel / formatDateRange fails when no temporal coverage is set


abulte:
When no temporal coverage is set on a dataset, the `InformationPanel` will crash because [`formatDateRange` expects `temporal_coverage.start` to be set](https://github.com/datagouv/udata-front/pull/411/files#diff-f7c70a99a67f7596b3df9db02c864dcc5e94a7fec923e84904b9a9d82385dd35R113).

Happens since #411 (datagouv-component 2.0.6)

PR #669, opened 2025-03-04
Fix error with dataset without temporal coverage

nicolaskempf57:
Fix https://github.com/datagouv/udata-front/issues/668


maudetes (approved):
:pray:

PR #682, opened 2025-04-10
Update links to new admin

ThibaudDauce:


udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/ResourceAccordion.vue
nicolaskempf57, line 1: I had to change l.99 to make it work.

```
const newResourceAdminUrl = new URL(config.admin_root + "community-resource/new/");
```

ThibaudDauce: What change did you make?

nicolaskempf57: I found what was my issue. The current code is working with a relative path like `beta/admin/` but it failed with an absolute path like `http://dev.local:3000/en/beta/admin/`

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/EditButton.vue
```diff
@@ -9,20 +9,36 @@
...
 const { t } = useI18n();
-const resourceType = props.isCommunityResource ? 'community-resource' : 'resource';
-const adminUrl = `${config.admin_root}dataset/${props.datasetId}/${resourceType}/${props.resourceId}`;
+
+const adminUrl = computed(() => {
+  if (config.new_admin) {
+    if ('owner' in props.resource && props.resource.owner) {
+      return `${config.admin_root}users/${props.resource.owner.id}/community-resources/#${props.resource.id}`
+    } else if ('organization' in props.resource && props.resource.organization) {
+      return `${config.admin_root}organization/${props.resource.organization.id}/community-resources/#${props.resource.id}`
```
maudetes, line 34: Ideally, having a community resource link without the user or organization in path (as in the legacy admin) would make it easier (and it would be aligned with other objects), right?

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/ResourceAccordion.vue
maudetes: I have the same issue with my `NEW_ADMIN_URL` locally

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/EditButton.vue
ThibaudDauce: It's a little bit harder because it's a modal and not a full page… But I can make it work…

maudetes: It's not necessarily worth the trouble now, it may be done when we migrate the datasets page?

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/ResourceAccordion.vue
ThibaudDauce: In the infra repo, it's set as `udata__new_admin_url: '/beta/admin/'` So relative? But I'm a little bit loss between `NEW_ADMIN_ROOT` from the Python and `VITE_ADMIN_ROOT` from the Vite build…

nicolaskempf57: The relative path is working for our repo infra because the same domain is used for both. Maybe we can handle both.

`VITE_ADMIN_ROOT` is the root to the old admin and not the new one.

ThibaudDauce: Ok, I didn't try to make it worked with different domains, I only checked it linked to :7000/beta/admin (which is a 404 in local env). Does it work with relative path for you?

udata_front/theme/gouvfr/templates/macros/metadata.html
```diff
@@ -20,7 +20,8 @@
 <meta name="static-root" content="{{ static('', _burst=False) }}" />
 <meta name="api-root" content="{{ url_for('api.root') }}" />
 <meta name="api-2-root" content="{{ url_for('apiv2.root') }}" />
-<meta name="admin-root" content="{{ url_for('admin.index') }}" />
+<meta name="admin-root" content="{{ beta_admin_url_for(path='/', fallback_path='/') }}" />
```
maudetes, line 23: I think fallback_path shouldn't have a `/` to prevent duplicate `//`
```suggestion
<meta name="admin-root" content="{{ beta_admin_url_for(path='/', fallback_path='') }}" />
```

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/EditButton.vue
ThibaudDauce: Ok

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/ResourceAccordion.vue
ThibaudDauce: What bothers me it's that I didn't change this part? It wasn't working before but I can change the 
`const newResourceAdminUrl = new URL(document.location.origin + config.admin_root + "community-resource/new/");` to `const newResourceAdminUrl = new URL(config.admin_root + "community-resource/new/");`

ThibaudDauce: Done in 7706277e

maudetes: I think it doesn't work then without a `NEW_ADMIN_URL` set?

udata_front/theme/gouvfr/templates/macros/metadata.html
```diff
@@ -20,7 +20,8 @@
 <meta name="static-root" content="{{ static('', _burst=False) }}" />
 <meta name="api-root" content="{{ url_for('api.root') }}" />
 <meta name="api-2-root" content="{{ url_for('apiv2.root') }}" />
-<meta name="admin-root" content="{{ url_for('admin.index') }}" />
+<meta name="admin-root" content="{{ beta_admin_url_for(path='/', fallback_path='') }}" />
+<meta name="new-admin" content="{{ config.NEW_ADMIN_URL }}" />
```
nicolaskempf57, line 24:
```suggestion
<meta name="new-admin-url" content="{{ config.NEW_ADMIN_URL }}" />
```

udata_front/theme/gouvfr/templates/macros/metadata.html
```diff
@@ -20,7 +20,8 @@
 <meta name="static-root" content="{{ static('', _burst=False) }}" />
 <meta name="api-root" content="{{ url_for('api.root') }}" />
 <meta name="api-2-root" content="{{ url_for('apiv2.root') }}" />
-<meta name="admin-root" content="{{ url_for('admin.index') }}" />
+<meta name="admin-root" content="{{ beta_admin_url_for(path='/', fallback_path='') }}" />
```
nicolaskempf57, line 23: Do we want to change this one too ? My understanding was that this one would still point to the old admin for the other reusers only.

We can't have the same links to the old and new admins because the old one uses singular and the new one uses plural.

ThibaudDauce: It will point to the old admin if there is no `NEW_ADMIN_URL`

udata_front/theme/gouvfr/templates/macros/metadata.html
ThibaudDauce: Done in [176ac2e](https://github.com/datagouv/udata-front/pull/682/commits/176ac2e8797a4193bcfe122e55d7b609a551c17a)

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/EditButton.vue
```diff
@@ -9,20 +9,35 @@
...
 const props = withDefaults(defineProps<Props>(), {
   isCommunityResource: false,
 });
 
 const { t } = useI18n();
-const resourceType = props.isCommunityResource ? 'community-resource' : 'resource';
-const adminUrl = `${config.admin_root}dataset/${props.datasetId}/${resourceType}/${props.resourceId}`;
+
+const adminUrl = computed(() => {
+  if (config.new_admin) {
```
maudetes, line 30: Shouldn't it be `new_admin_url`?

maudetes (commented):
I'm confused between the double information between metadata.html and datagouv components .env.
We don't need the second one right?

ThibaudDauce:
> I'm confused between the double information between metadata.html and datagouv components .env. We don't need the second one right?

No I think we don't need the first one (the one in metadata, because it's only for Vue component inside udata-front (and most of udata-front is just `beta_admin_url_for` in Python with no links in Vue). The second one in `.env` is for datagouv components (like ResourceAccordion which has links)

udata_front/theme/gouvfr/datagouv-components/src/components/ResourceAccordion/EditButton.vue
ThibaudDauce: No it's a boolean indicating if we are using the new type of URLs (with the plural)

ThibaudDauce: If `VITE_NEW_ADMIN` is not set, we use the old URLs (for other users of udata/udata-front), but if we set it to `=true` we will use the plurals and the new links

ThibaudDauce: I understand your point, I added a default value to false in 64ed7088

maudetes: and then to `true` in https://github.com/datagouv/udata-front/pull/682/commits/b578a4e322a850212cfefbb17a8388c58dd12dc4 :ok_hand:

maudetes (approved):

PR #685, opened 2025-04-15
Use csv queryset instead of .visible for csv routes

maudetes:
We should use the same queryset as in exported catalogs built by `export-csv` job.

Indeed, now we don't have the same visibility rules, excluding archived datasets in `/datasets.csv` route whereas they are included in the daily exported catalog.


Pierlou (approved):
Nice, thanks for the quick fix! 🙏

PR #688, opened 2025-04-23
Remove MAAF backend

ThibaudDauce:
Require https://github.com/opendatateam/udata/pull/3300


maudetes (approved):
:pray: thank you!

I'm still surprised that somehow it's still working in production for all those years.

